PR #2, opened 2014-09-20
Expose underlying data for BigUint and BigInt.

chromaticbum:
It would be useful to have access to the underlying data used to represent these numbers for use in certain algorithms that can be optimized with access to the underlying data as seen in:
ftp://ftp.cs.pdx.edu/smn/ipsec/skip/bnlib/jacobi.c

Perhaps another way to go about this would be to implement the Index trait for these two data types.


gsingh93:
This should probably be closed in favor of #54.

PR #46, opened 2015-01-12
Replace deprecated int references with i32/isize.

jbcrail:
I replaced most int refs with i32, and I used isize to hold the number
of default digits.


alexcrichton:
Thanks!

PR #47, opened 2015-01-12
Remove range function.

jbcrail:
For the pi benchmark, I replaced the range function with the new range
syntax.


alexcrichton:
Thanks!

PR #48, opened 2015-01-18
Utilize range x..y sugar

frewsxcv:


alexcrichton:
Unfortunately this doesn't currently compile (see https://github.com/rust-lang/rust/issues/20811 for details), so I'm going to close this for now. Feel free to resubmit though once that bug is fixed!

PR #52, opened 2015-01-20
Rename BigDigit module to big_digit

renato-zannon:
Without this rename, the library fails to compile on Rust master, since rust-lang/rust#21269 has landed.


alexcrichton:
Thanks!

PR #53, opened 2015-01-20
Fixed some typos and added some examples

gsingh93:


gsingh93:
Wait, that "be" in that comment wouldn't happen to mean "big endian", would it?

gsingh93:
Oops, I committed more than I meant to commit. Fixing now.

huonw:
I suspect it is meant to be big-endian; could you change it to be the long form? (it's certainly highly confusing as written.)

huonw:
Hm, actually, aren't these being stored as little-endian?

huonw:
Also, the example you added needs some imports to compile (see the travis): `use num::{BigInt, ToBigInt};`.

(Sorry for the comment spam.)

gsingh93:
Yea, I believe it is little-endian. To make sure, I ran:
```
println!("{:?}", BigInt::new(Sign::Plus, vec![1, 2]));
```
which gave `8589934593`, meaning the leftmost digit in the vector is the least significant digit. I'll change the comment to say "little-endian" then.

I also realized I changed this for `BigUint` thinking it was a `BigInt`. I've updated the documentation for both of them now.

alexcrichton:
Thanks!

PR #54, opened 2015-01-21
Implemented from_bytes and to_bytes

gsingh93:
This commit implements `from_bytes` and `to_bytes` methods for `BigInt` and `BigUint`. These methods are very useful when working with BigInt types.

The current code works with big-endian slices/vectors, while the existing code (`new()` and `from_slice()`) uses little-endian slices/vectors. My opinion is that big-endian is preferable, but I can change this if necessary.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -56,19 +56,21 @@
 //! println!("{}", a * b);
 //! ```
 
+extern crate "rustc-serialize" as rustc_serialize;
+
 use Integer;
 
 use std::default::Default;
 use std::iter::repeat;
-use std::num::FromStrRadix;
-use std::num::{Int, ToPrimitive, FromPrimitive};
+use std::num::{Int, ToPrimitive, FromPrimitive, FromStrRadix};
 use std::ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Neg, Rem, Shl, Shr, Sub};
 use std::rand::Rng;
 use std::str::{self, FromStr};
-use std::{cmp, fmt, hash};
-use std::cmp::Ordering;
-use std::cmp::Ordering::{Less, Greater, Equal};
+use std::{cmp, fmt, hash, mem};
+use std::cmp::Ordering::{self, Less, Greater, Equal};
 use std::{i64, u64};
+
+use rustc_serialize::hex::ToHex;
 
 use {Num, Unsigned, CheckedAdd, CheckedSub, CheckedMul, CheckedDiv, Signed, Zero, One};
 use self::Sign::{Minus, NoSign, Plus};
@@ -874,6 +876,75 @@ impl BigUint {
 
     /// Creates and initializes a `BigUint`.
     ///
+    /// The bytes are in big-endian byte order.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num::bigint::BigUint;
+    ///
+    /// assert_eq!(BigUint::from_bytes_be("A".as_bytes()),
+    ///            BigUint::parse_bytes("65".as_bytes(), 10).unwrap());
+    /// assert_eq!(BigUint::from_bytes_be("AA".as_bytes()),
+    ///            BigUint::parse_bytes("16705".as_bytes(), 10).unwrap());
+    /// assert_eq!(BigUint::from_bytes_be("AB".as_bytes()),
+    ///            BigUint::parse_bytes("16706".as_bytes(), 10).unwrap());
+    /// assert_eq!(BigUint::from_bytes_be("Hello world!".as_bytes()),
+    ///            BigUint::parse_bytes("22405534230753963835153736737".as_bytes(), 10).unwrap());
+    /// ```
+    #[inline]
+    pub fn from_bytes_be(bytes: &[u8]) -> BigUint {
+        if bytes.is_empty() {
+            Zero::zero()
+        } else {
+            BigUint::parse_bytes(bytes.to_hex().as_bytes(), 16).unwrap()
+        }
+    }
+
+    /// Creates and initializes a `BigUint`.
+    ///
+    /// The bytes are in little-endian byte order.
+    #[inline]
+    pub fn from_bytes_le(bytes: &[u8]) -> BigUint {
+        let mut v = bytes.to_vec();
+        v.reverse();
+        BigUint::from_bytes_be(&*v)
+    }
+
+    /// Returns the byte representation of the `BigUint` in little-endian byte order.
+    #[inline]
+    pub fn to_bytes_le(&self) -> Vec<u8> {
+        let mut result = Vec::new();
+        for word in self.data.iter() {
+            let mut w = *word;
+            for _ in 0..mem::size_of::<BigDigit>() {
+                let b = (w & 0xFF) as u8;
+                w = w >> 8;
+                result.push(b);
+            }
+        }
+
+        if let Some(index) = result.iter().rposition(|x| *x != 0) {
+            result.truncate(index + 1);
+        }
+
+        if result.is_empty() {
+            vec![0]
+        } else {
+            result
+        }
+    }
+
+    /// Returns the byte representation of the `BigUint` in big-endian byte order.
+    #[inline]
+    pub fn to_bytes_be(&self) -> Vec<u8> {
+        let mut v = self.to_bytes_le();
+        v.reverse();
+        v
+    }
+
+    /// Creates and initializes a `BigUint`.
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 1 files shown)


huonw:
I'm not sure I understand the value of these, could you go into more detail about what they're useful for?

gsingh93:
My particular use case is in cryptography, converting between the bytes of messages to integers that I can do certain mathematical operations on.

But there are lots of other uses, there's even an [issue](https://github.com/rust-lang/num/issues/5) for another use case, and an [unmerged pull request](https://github.com/rust-lang/num/pull/2) that exposes the underlying data (which isn't the same thing, but it shows people want any easy way to access the bytes of the data).

gsingh93:
I was using this version in one of my projects, and I realized that there's no reason `from_bytes` should ever fail. I should be able to call `unwrap()` within the function and return a `BigInt` if the other functions are implemented correctly. If you decide to merge this change, I can go ahead and make that change here as well if you agree.

huonw:
It sounds like there is value for this, although I wonder if the functions could be called `to_bytes_be` and `from_bytes_be`.

(I've also asked @alexcrichton for feedback, who is likely to have better ideas than me.)

src/bigint.rs
```diff
@@ -878,6 +882,45 @@ impl BigUint {
...
+
+    /// Returns the byte representation of the `BigUint`
+    ///
+    /// The bytes are in big-endian byte order
+    #[inline]
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut a = self.clone();
+        let mut result = Vec::new();
+        let byte_size = ToBigUint::to_biguint(&256).unwrap();
+        while !a.is_zero() {
```
huonw: I think this can be far more efficient, at the very least it could use shifts and masks instead of division and modulus, but, even better: 256 divides the base (2<sup>32</sup>), so the conversion can be done word-by-word, by iterating over `self.data` (in reverse, I guess).

src/bigint.rs
```diff
@@ -1576,6 +1619,49 @@ impl BigInt {
...
+            Some(b) => Some(BigInt::from_biguint(sign, b)),
+            None => None
+        }
+    }
+
+    /// Returns the byte representation of the `BigInt`
+    ///
+    /// The bytes are in big-endian byte order
+    #[inline]
+    pub fn to_bytes(&self) -> Vec<u8> {
```
huonw: Could this call `BigUint`s equivalent via `self.data.to_bytes()`?

gsingh93:
I've made some of the mentioned changes. I still have to write `from_bytes_le`, add some more tests for the new functions added, and possibly add a few more examples, but I wanted to get some feedback on the changes and ask a few questions:

1. I noticed that some methods in `BigInt` that rely on `BigUint` aren't tested as long as the corresponding method in `BigUint` is tested (i.e. `from_slice`). Is this intentional or an oversight? If my methods in `BigInt` only rely on `BigUint`, should I only test `BigUint`?
2. I'd really like to make `from_bytes_be` not return an option. I thought this was possible, but the one case where it could fail is if an empty slice is passed in. Should we assume that corresponds to the number 0, or should we return an option?
3. I don't see the benefit of having `BigInt::new()` take a vector of `BigDigit`s. This seems inconvenient to me (although I haven't been using the API for too long, maybe I'm wrong). I think it should be switched to use either the `from_bytes_le` or `from_bytes_be` functions.

src/bigint.rs
```diff
@@ -879,6 +883,65 @@ impl BigUint {
...
+        BigUint::parse_bytes(msg.to_hex().as_bytes(), 16).unwrap()
+    }
+
+    /// Creates and initializes a `BigUint`.
+    ///
+    /// The bytes are in little-endian byte order
+    #[inline]
+    pub fn from_bytes_le(msg: &[u8]) -> BigUint {
+        let mut s = msg.to_hex();
+        unsafe {
```
gsingh93: Is there a safe way to do this (without any extra allocations)?

block8437:
I'm currently working on an implementation of the Socialist Millionaire (a cryptographic protocol in order to determine if shared secrets are equal) and it is not possible for me to do that without the ability to send BigNums over the network. Please consider merging this as soon as possible. Thanks.

gsingh93:
@alexcrichton, @huonw, any feedback?

src/bigint.rs
huonw: Use `.into_bytes()` to consume the `String` into a `Vec<u8>`.

src/bigint.rs
```diff
@@ -879,6 +883,65 @@ impl BigUint {
...
+        BigUint::parse_bytes(s.as_bytes(), 16).unwrap()
+    }
+
+    /// Returns the byte representation of the `BigUint` in little-endian byte order.
+    #[inline]
+    pub fn to_bytes_le(&self) -> Vec<u8> {
+        let mut result = Vec::new();
+        for word in self.data.iter() {
+            let mut w = *word;
+            while w != 0 {
```
huonw: I suspect this isn't quite right, e.g. (with u16 words), the number `0x1_0002` is represented as `[0002u16, 0001u16]` and the bytes should be `[2, 0, 1]`, but this will be triggered too early and the bytes will be `[2, 1]`.

Maybe just a fixed length loop would work best: `for _ in 0..mem::size_of::<BigDigit>() { ... }`.

That said, this truncating behaviour is probably desirable for the last word (e.g. it would be nice for the example above to not be `[2, 0, 1, 0]`).

huonw: (Another special case: should `0.to_bytes_le()` give `[0]` or `[]`? I'm inclined to say the former, even though that isn't the shortest representation that `from_bytes_le` will parse.)

huonw:
> I noticed that some methods in BigInt that rely on BigUint aren't tested as long as the corresponding method in BigUint is tested (i.e. from_slice). Is this intentional or an oversight? If my methods in BigInt only rely on BigUint, should I only test BigUint?

It'd be nice to at least have a few tests (e.g. checking that the `sign` argument does the right thing).

> I'd really like to make from_bytes_be not return an option. I thought this was possible, but the one case where it could fail is if an empty slice is passed in. Should we assume that corresponds to the number 0, or should we return an option?

Returning 0 is fine by me.

> I don't see the benefit of having BigInt::new() take a vector of BigDigits. This seems inconvenient to me (although I haven't been using the API for too long, maybe I'm wrong). I think it should be switched to use either the from_bytes_le or from_bytes_be functions.

I'm not sure, it doesn't seem hugely useful; maybe we could just rename it and not have a `new` function at all (i.e. everything is `from_*`).

alexcrichton:
From an API level this seems fine to me, and it looks like @huonw has the implementation covered :)

gsingh93:
Ok, I've addressed those issues and squashed all the commits. Let me know if you find any more problems.

gsingh93:
Any more comments?

huonw:
Will take a (probably) final look tomorrow. (Feel free to ping me here or on IRC if you don't get a response in the next 24 hours.)

src/bigint.rs
```diff
@@ -874,6 +876,75 @@ impl BigUint {
...
+    #[inline]
+    pub fn from_bytes_le(bytes: &[u8]) -> BigUint {
+        let mut v = bytes.to_vec();
+        v.reverse();
+        BigUint::from_bytes_be(&*v)
+    }
+
+    /// Returns the byte representation of the `BigUint` in little-endian byte order.
+    #[inline]
+    pub fn to_bytes_le(&self) -> Vec<u8> {
```
huonw: Possible extension: make this return an `Iterator` rather than returning a `Vec`.

src/bigint.rs
```diff
@@ -1578,6 +1649,49 @@ impl BigInt {
...
+
+    /// Creates and initializes a `BigInt`.
+    ///
+    /// The bytes are in little-endian byte order
+    #[inline]
+    pub fn from_bytes_le(sign: Sign, bytes: &[u8]) -> BigInt {
+        BigInt::from_biguint(sign, BigUint::from_bytes_le(bytes))
+    }
+
+    /// Returns the byte representation of the `BigInt` in little-endian byte order
```
huonw: Could you document what this does for negative numbers? (And below.)

I'm inclined to think it should require that `self` is positive? (with an `assert!`, maybe?)

src/bigint.rs
huonw: (I don't think this needs to be done now.)

src/bigint.rs
gsingh93: Actually, I'd prefer this to simply return the byte representation regardless of the sign. If the user wants the underlying bytes representing the integer, they should be allowed to get them. If they need to handle negative numbers a certain way, they can still access that information from the original `BigInt` and handle that case accordingly.

gsingh93: Does this sound good, @huonw?

huonw: Sounds reasonable. Could you add this to the documentation?

Hm, maybe it could return `(Sign, Vec<u8>)` to be a true inverse to `from_bytes_le`. One only needs to write an additional `.1` to access the `Vec<u8>` itself.

gsingh93: I like the inverse idea. I've just updated that and mentioned it in the documentation.

src/bigint.rs
```diff
@@ -874,6 +876,75 @@ impl BigUint {
 
     /// Creates and initializes a `BigUint`.
     ///
+    /// The bytes are in big-endian byte order
```
huonw: This (and a few other sentences below) should have a trailing `.`. Other than that, I think it's ready.

gsingh93: Fixed and squashed.

huonw:
Thanks!

PR #58, opened 2015-02-03
Update to build with Rust 1.0.0-nightly

mbrubeck:
This is a minimal set of changes to fix build errors and warnings.  More work could be done to convert other methods from Option to Result, and to use more error chaining.


alexcrichton:
Thanks!

PR #60, opened 2015-02-19
Change `radix` to u32, to build with new rustc.

arbitrary-cat:


arbitrary-cat:
The build fails due to `rustc: command not found` and similar for cargo. It builds fine on my system with the compiler installed =P.

PR #65, opened 2015-03-07
pow: avoid unnecessary overflows

cuviper:
The code was performing an extra squaring of the base, which might
trigger an arithmetic overflow that doesn't matter to the result.  Now
this squaring is only attempted when enough exp remains to need it.

A new doctest tries pow(6u8, 3), where an extra square would exceed 256.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,7 +121,8 @@ pub fn abs_sub<T: Signed>(x: T, y: T) -> T {
 /// ```rust
 /// use num;
 ///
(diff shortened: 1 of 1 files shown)


cuviper:
FWIW, the standard library has a related change in `Int::pow`, rust-lang/rust@c8db89aa82573b89481fde598da6e54371f266cb, but IMO that approach was needlessly complex.  But we don't have generic `overflowing_mul` available here anyway.

alexcrichton:
Thanks!

PR #75, opened 2015-04-03
Vendor deprecated/unstable traits from std::num

### LINKED ISSUE #74: Will `std::num::Float` be moved here?
See https://github.com/PistonDevelopers/piston/issues/864

We use the `Float` for generics in some piston libraries, and wonder what's the plan of replacing the trait.

aturon:
This commit brings in a load of unstable and/or deprecated traits from
the `std::num` module. These traits provide for some degree of generic
programming over numeric types. They are not stable in `std` mostly
because we want more time to iterate on their design. Moving them to the
`num` crate allows existing code to keep using this functionality as we
do so.

r? @alexcrichton 

Closes #74

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -11,10 +11,13 @@
 //! Numeric traits for generic mathematics
 
 use std::intrinsics;
-use std::ops::{Add, Div, Mul, Neg, Rem, Sub};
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

PR #76, opened 2015-04-03
Update to 1.0.0-beta (stable only)

aturon:
Along the way, this commit also rationalizes the traits a bit more.

Moving to stable required vendoring a couple of minor things (notably,
radix formatting).

r? @alexcrichton

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -62,12 +62,10 @@ extern crate rustc_serialize;
 
 use Integer;
 
-use core::num::ParseIntError;
-
 use std::default::Default;
 use std::error::Error;
 use std::iter::repeat;
-use std::num::{Int, ToPrimitive, FromPrimitive, FromStrRadix};
+use std::num::ParseIntError;
 use std::ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Neg, Rem, Shl, Shr, Sub};
 use std::str::{self, FromStr};
 use std::{cmp, fmt, hash, mem};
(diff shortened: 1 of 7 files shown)


andersk:
This makes it impossible to construct a `Ratio` of unsigned integers; is that fixable?

andersk:
In particular, if we reimplement unary negation in terms of subtraction from `Zero::zero()` (which will be dead code in the unsigned case), then we can get rid of all `T: Neg` constraints except the ones on `Neg` and `Signed`.

```patch
@@ -98,9 +98,10 @@ impl<T: Clone + Integer + PartialOrd + Neg<Output = T>> Ratio<T> {
         self.denom = self.denom.clone() / g;
 
         // keep denom positive!
-        if self.denom < Zero::zero() {
-            self.numer = -self.numer.clone();
-            self.denom = -self.denom.clone();
+        let zero: T = Zero::zero();
+        if self.denom < zero.clone() {
+            self.numer = zero.clone() - self.numer.clone();
+            self.denom = zero.clone() - self.denom.clone();
         }
     }
 
@@ -146,7 +147,13 @@ impl<T: Clone + Integer + PartialOrd + Neg<Output = T>> Ratio<T> {
         let two: T = one.clone() + one.clone();
 
         // Find unsigned fractional part of rational number
-        let fractional = self.fract().abs();
+        let signed_fractional = self.fract();
+        let fractional = if fractional < Zero::zero() {
+            let zero: Ratio<T> = Zero::zero();
+            zero - signed_fractional
+        } else {
+            signed_fractional
+        };
 
         // The algorithm compares the unsigned fractional part with 1/2, that
         // is, a/b >= 1/2, or a >= b/2. For odd denominators, we use
```

alexcrichton:
Nice! I think @andersk has a good suggestion, but other than it looks great to me!

aturon:
@andersk Thanks for the suggestion! I've updated the PR accordingly.

@alexcrichton re-r?

andersk:
On the other hand, I’m not sure it makes sense to omit the `Neg` requirement for `Complex` `conj` and `inv`, because the negation in those methods is _not_ dead code for unsigned types. I opened #77 for that.

PR #77, opened 2015-04-03
Require Neg for Complex conj and inv

andersk:
If `T` is an unsigned integer type, these methods are guaranteed to overflow unless the result is actually real, so we should disallow them for the same reason that `Neg` was removed from these types.


alexcrichton:
Thanks!

PR #81, opened 2015-04-12
Export the #[derive(FromPrimitive)] from rust proper

erickt:
This changes the name to `#[derive(NumFromPrimitive)]` in order to not conflict with the libsyntax macro.

If the general direction of this num library will supplant `std::num`, then I'd instead prefer to do the same trick we did for rustc-serialize, and have the compiler add a `#[derive(RustcFromPrimitive)]` that uses the trait in this crate instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
-/target
 /Cargo.lock
+/num_macros/Cargo.lock
+/num_macros/target
+/target
--- a/num_macros/Cargo.toml
+++ b/num_macros/Cargo.toml
@@ -1,6 +1,5 @@
 [package]
-
-name = "num"
+name = "num_macros"
 version = "0.1.22"
 authors = ["The Rust Project Developers"]
 license = "MIT/Apache-2.0"
@@ -9,14 +8,9 @@ repository = "https://github.com/rust-lang/num"
 documentation = "http://doc.rust-lang.org/num"
 keywords = ["mathematics", "numerics"]
 description = """
-Simple numerics. This crate contains basic arbitrary-sized integer,
-rational, and complex types.
+Numeric syntax extensions.
 """
 
-[dependencies]
-rustc-serialize = "0.3.5"
-rand = "0.3.3"
-
-[dev-dependencies.num_macros]
-path = "num_macros/"
-version = "0.1.22"
+[lib]
+name = "num_macros"
+plugin = true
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -16,3 +16,7 @@ rational, and complex types.
 [dependencies]
 rustc-serialize = "0.3.5"
 rand = "0.3.3"
+
+[dev-dependencies.num_macros]
+path = "num_macros/"
+version = "0.1.22"
```
alexcrichton: I think it's fine to start this at `0.1.0`, and could this be formatted as:

```toml
[dev-dependencies]
num-macros = { path = "num-macros", version = "0.1.0" }
```

num_macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "num_macros"
```
alexcrichton: Alluded to above, but this package should be called `num-macros` with our current convention of crate names on crates.io

num_macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.22"
+authors = ["The Rust Project Developers"]
+license = "MIT/Apache-2.0"
+homepage = "https://github.com/rust-lang/num"
+repository = "https://github.com/rust-lang/num"
+documentation = "http://doc.rust-lang.org/num"
+keywords = ["mathematics", "numerics"]
+description = """
+Numeric syntax extensions.
```
alexcrichton: Could this description mention `derive(NumFromPrimitive)`?

.gitignore
```diff
@@ -1,2 +1,4 @@
-/target
 /Cargo.lock
+/num_macros/Cargo.lock
+/num_macros/target
+/target
```
alexcrichton: I think it's also fine to just remove the leading `/` from the first two entries

alexcrichton:
Nice! Could you also edit `.travis.yml` to run the tests for this new crate as well?

Also cc @aturon, how do you feel about this?

aturon:
Sorry I missed this before. Someone wrote a [macro](https://github.com/rust-lang/rust/issues/16920#issuecomment-90451501) to replace this derive mode, so my preference for the moment would be to remove this altogether.

The rust-lang `num` crate may eventually be a place to test ideas for `std::num`, but at the moment it's largely a place that deprecated items have been dropped into.

nwin:
The macro solution is quite limited. It is difficult to impossible to document your enums (#24189). I would be very much in favor of `#[derive(NumFromPrimitive)]` as an immediate solution.

aturon:
OK, taking a closer look at this, I'm fine with this direction for the moment. (In the long run, I still hope we can provide better built-in support in Rust itself for this kind of conversion of enums).

alexcrichton:
I renamed `num_macros` to `num-macros` and otherwise just tidied up the merge a bit, thanks again @erickt!

PR #82, opened 2015-04-15
Improve bigint docs.

nham:
This commit converts instances of `<string literal>.as_bytes()` in the bigint doc examples to byte string literals. It also adds examples for `from_bytes_le()` and `to_bytes_be()`.


alexcrichton:
Thanks!

PR #85, opened 2015-04-22
Removed deprecation attributes that can no longer be used outside of std

### LINKED ISSUE #84: Compilation is broken because of stability attributes
Stability attributes are now errors in the nightlies, so num 0.1.22 no longer compiles.

I don't have time to fix it right now, so I'll just open an issue in case no-one is already fixing it.

mitchmindtree:
Also incremented the crates.io version ready for the next `cargo publish`

This closes #84

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "num"
-version = "0.1.22"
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #86, opened 2015-04-23
version bump

steveklabnik:


alexcrichton:
Hm this library currently doesn't build so unfortunately we can't publish a new version.

steveklabnik:
I swear it built for me... i think multirust is messing with me.

PR #87, opened 2015-04-23
Use latest rustc-serialize and rand, use correct version of from_str_radix, increment version for crates.io

### LINKED ISSUE #88: Doesn't compile as a dependency of a dependency
I can't compile my crate because a dependency uses `num` which doesn't compile

```
53:37 error: use of unstable library feature 'float_from_str_radix': recently moved API
             <$t>::from_str_radix(s, radix)
                   ^~~~~~~~~~~~~~
57:2 note: in expansion of float_trait_impl!
60:36 note: expansion site
 53:37 help: add #![feature(float_from_str_radix)] to the crate attributes to enable
 53:37 error: use of unstable library feature 'float_from_str_radix': recently moved API
             <$t>::from_str_radix(s, radix)
                   ^~~~~~~~~~~~~~
57:2 note: in expansion of float_trait_impl!
60:36 note: expansion site
 53:37 help: add #![feature(float_from_str_radix)] to the crate attributes to enable
```

mitchmindtree:
Closes #86 

Edit: also closes #88

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -43,14 +43,160 @@ macro_rules! int_trait_impl {
     )*)
 }
 
+// FIXME: Temporary replacements for unstable ::std::num::ParseFloatError and
+// ::std::num::FloatErrorKind. These can be removed once the std float implementation of
+// from_str_radix stabilises.
+pub enum FloatErrorKind { Empty, Invalid }
+pub struct ParseFloatError { pub kind: FloatErrorKind }
+
+// FIXME: This should be removed and replaced with the std implementation of from_str_radix once
+// it is stabilised.
 macro_rules! float_trait_impl {
     ($name:ident for $($t:ty)*) => ($(
         impl $name for $t {
-            type FromStrRadixErr = ::std::num::ParseFloatError;
-            fn from_str_radix(s: &str, radix: u32)
-                              -> Result<Self, ::std::num::ParseFloatError>
+            type FromStrRadixErr = ParseFloatError;
+            fn from_str_radix(src: &str, radix: u32)
+                              -> Result<Self, ParseFloatError>
             {
-                Num::from_str_radix(s, radix)
+                use self::FloatErrorKind::*;
+                use self::ParseFloatError as PFE;
+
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Hmm trav seems to be erroring on [this line](https://github.com/rust-lang/num/blob/master/src/traits.rs#L53), going to update to the latest nightly and see if I can reproduce the error on my end and fix it.

mitchmindtree:
The issue is that the `.from_str_radix` method has been "moved" - the docs [mention this](http://doc.rust-lang.org/std/primitive.f64.html#method.from_str_radix) but they do not mention where it has been moved too.

mitchmindtree:
#88 is related

mitchmindtree:
The issue was that rustc was choosing the recently re-implemented, unstable version of the `from_str_radix` method rather than the local `Num` implementation. Using `Num` on the culprit line seems to have fixed it.

src/traits.rs
```diff
@@ -50,7 +50,7 @@ macro_rules! float_trait_impl {
             fn from_str_radix(s: &str, radix: u32)
                               -> Result<Self, ::std::num::ParseFloatError>
             {
-                <$t>::from_str_radix(s, radix)
+                Num::from_str_radix(s, radix)
```
alexcrichton: Isn't this just an infinite recursion? (e.g. this looks like it's just calling itself)

mitchmindtree: Ahh you're right, my bad (getting late here).

Any idea why this is only a problem for the float impl of Num and not the integer impl?

alexcrichton: The float implementations are currently unstable in the standard library unfortunately.

mitchmindtree:
Ok, as alex pointed out my last attempt was just causing an infinite loop.

This last commit copies the implementation from std into the Num implementation as a temporary avoidance of using the unstable feature. I've left a `FIXME` indicating it should be replaced once the std `from_str_radix` is stabilised.

@alexcrichton is this ok? (currently waiting on travis)

PR #89, opened 2015-04-25
traits.rs: Implement Bounded for tuples of Bounded types

joshtriplett:
For example, the following will print (4294967295, 65535):

extern crate num;
use num::traits::Bounded;

fn main() {
    let t : (u32, u16) = Bounded::max_value();
    println!("{:?}", t);
}


alexcrichton:
Thanks!

PR #96, opened 2015-05-18
Update num-macros to rust master

renato-zannon:
Now compiling on `rustc 1.2.0-nightly (0cc99f9cc 2015-05-17) (built 2015-05-18)`


alexcrichton:
Thanks!

PR #98, opened 2015-05-24
Update num-macros to rust nightly

clatour:
Now compiles with `rustc 1.2.0-nightly (0fc0476e6 2015-05-24) (built 2015-05-23)`


alexcrichton:
Thanks!

PR #101, opened 2015-06-02
Make rustc-serialize, bignum, rational, complex into opt-out features

### LINKED ISSUE #63: Consider splitting into smaller crates
Given that the data types in here are relatively “fundamental” (some having even been part of the standard library in the past), they're likely to be used in a lot of places. It would therefore be nice to have more fine-grained control over these dependencies, to minimize unnecessary bloat. In particular, `rational`, `bigint`, and `complex` seem like reasonably separate concepts/libraries.

For example, I was just now surprised to see that adding a simple `Rational` to my code pulled in dependencies relating to logging, serialization, and random numbers. I want to use a common type that consumers of my library are likely to be able to use without converting from my own `Rational` type, but I also don't want my code to transitively pull in a lot of stuff for its users.

bluss:
Make rustc-serialize, bignum, rational, complex into opt-out features

Making bignum optional allows skipping the rustc-serialize and rand
dependencies too, and it makes a big difference in num's build time.

With default (all) features, clean build time including dependencies: 27
seconds.

With no default features, clean build time including dependencies (none):
5 seconds.

Fixes #63

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,15 @@ rational, and complex types.
 """
 
 [dependencies]
-rustc-serialize = "0.3.13"
-rand = "0.3.8"
+rustc-serialize = { version = "0.3.13", optional = true }
+rand = { version = "0.3.8", optional = true }
+
+[features]
+
+complex = []
+rational = []
+bigint = ["rustc-serialize", "rand"]
+default = ["complex", "rational", "bigint"]
 
 [[bench]]
 name = "shootout-pidigits"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - |
+    (for feature in '' bigint rational complex; do
+      cargo test --verbose --no-default-features --features="$feature" || exit 1
+    done)
+  - |
     [ $TRAVIS_RUST_VERSION != nightly ] || (
       cargo bench &&
       cargo test --verbose --manifest-path=num-macros/Cargo.toml
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -58,8 +58,6 @@
 //! # }
 //! ```
 
-extern crate rustc_serialize;
-
 use Integer;
 
 use std::default::Default;
--- a/src/complex.rs
+++ b/src/complex.rs
@@ -20,7 +20,8 @@ use {Zero, One, Num, Float};
 // probably doesn't map to C's _Complex correctly.
 
 /// A complex number in Cartesian form.
-#[derive(PartialEq, Copy, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
+#[derive(PartialEq, Copy, Clone, Hash, Debug)]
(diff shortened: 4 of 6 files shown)


bluss:
5 vs 27 seconds of build time will have a big impact on how often this library is a viable dependency.

bluss:
This is one way to basically fix #63

alexcrichton:
I think this is an ok step to take for now, although I'd prefer to explore moving these to external crates in the future. I'd also like to take a look at the whole API of `num`, however, if we do that, so it makes sense to probably hold off on splitting things up until that point. Just having a small set of features doesn't hinder development too much.

Cargo.toml
```diff
@@ -13,9 +13,19 @@ Simple numerics. This crate contains basic arbitrary-sized integer,
 rational, and complex types.
 """
 
-[dependencies]
-rustc-serialize = "0.3.13"
-rand = "0.3.8"
+[dependencies.rustc-serialize]
+version = "0.3.13"
+optional = true
```
alexcrichton: Could these instead be written as:

```toml
[dependencies]
rustc-serialize = { version = "0.3.13", optional = true }
rand = { version = "0.3.8", optional = true }
```

Cargo.toml
```diff
@@ -13,9 +13,19 @@ Simple numerics. This crate contains basic arbitrary-sized integer,
...
+
+[dependencies.rand]
+version = "0.3.8"
+optional = true
+
+[features]
+
+rational = []
+bigint = ["rustc-serialize", "rand"]
+default = ["rational", "bigint"]
```
alexcrichton: Can you modify `.travis.yml` to test with features disabled and with only one of each individual feature enabled?

Cargo.toml
```diff
@@ -13,9 +13,19 @@ Simple numerics. This crate contains basic arbitrary-sized integer,
...
+version = "0.3.13"
+optional = true
+
+[dependencies.rand]
+version = "0.3.8"
+optional = true
+
+[features]
+
+rational = []
```
alexcrichton: How come `rational` has a feature, but `complex` doesn't? I'd expect the biggest share of compile time to be with bigints?

bluss: I'm not sure. Maybe rational doesn't need a feature, I didn't look how big it was. I expected complex to be small. We can feature gate them all?

bluss:
@alexcrichton I thought of splitting them first. Using features is easy, but is it unnecessary churn?  It does seem to be a perfect fit for cargo features.

Cargo.toml
bluss: This was a good idea, now I found more stuff to fix :)

bluss:
Ok now the split works. The main thing here is to be able to remove the rand, serialize dependencies optionally, but also the bigint code. We could do it with fewer feature flags.

alexcrichton:
>  It does seem to be a perfect fit for cargo features.

In the past the trend has been to split into separate crates, so I think this will probably take that direction eventually. I'm not sure if this is the greatest mapping to features, but it certainly works!

alexcrichton:
Thanks @bluss!

PR #104, opened 2015-07-03
Add function to compute the modular exponentiation for BigInts

kunerd:
Implementation of the 'Left-to-right k-ary exponentiation' algorithm.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -1855,6 +1855,58 @@ impl From<ParseIntError> for ParseBigIntError {
     }
 }
 
+const TABLE_BASE: usize = 5;
+
+pub trait ModPow<T> {
+    fn mod_pow(&self, exp: &T, m: &T) -> T;
+}
+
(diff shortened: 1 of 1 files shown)


kunerd:
Hi, this is my first pull request to a rust library. It definitely needs some documentation, but I thought it would be better to wait for some feedback first and add the documentation at the end.

alexcrichton:
Unfortunately I'm pretty unfamiliar with the implementation here, so it may be best to put this as an extension trait on crates.io instead of an extension trait in-tree.

kunerd:
It is an implementation of the LR-k-ary algorithm as described in the "[Handbook of Applied Cryptography](http://cacr.uwaterloo.ca/hac/about/chap14.pdf), 14.82".

Your idea to put this as an extension trait to crates.io is also OK for me.

alexcrichton:
Ok, let's take that strategy for now, thanks regardless for the PR though!

PR #112, opened 2015-09-07
Implement Stein's algorithm for gcd

Emerentius:
The binary gcd algorithm offers significant speedups when the CPU has a builtin instruction for counting trailing zeros.

I benched the improvements with random numbers. Code is at  https://github.com/Emerentius/gcd_bench . These are only rough measurements, I used 50 (Edit: Now 5000) number pairs for each. The old gcd liked to crash for `i8` when I used more.

These are the improvements on my machine (Intel i5 2500k). All numbers are averaged.

```
u8
gcd:            29.65 ns / call
binary_gcd:     20.70 ns / call (  43.2% faster )

u16
gcd:            58.66 ns / call
binary_gcd:     42.50 ns / call (  38.0% faster )

u32
gcd:            96.71 ns / call
binary_gcd:     72.13 ns / call (  34.1% faster )

u64
gcd:           314.31 ns / call
binary_gcd:    148.77 ns / call ( 111.3% faster )

i8
gcd:            28.31 ns / call
binary_gcd:     17.33 ns / call (  63.3% faster )

i16
gcd:            62.20 ns / call
binary_gcd:     37.74 ns / call (  64.8% faster )

i32
gcd:           102.19 ns / call
binary_gcd:     65.84 ns / call (  55.2% faster )

i64
gcd:           388.04 ns / call
binary_gcd:    136.96 ns / call ( 183.3% faster )
```

As a bonus: `i8::min_value() % -1` and variants panic in the implementation of the euclidean algorithm. This implementation doesn't panic. Like in the euclidean implementation, the special case `gcd(i8::min_value(), i8::min_value()) == i8::min_value()` (negative output) still exists. Edit: In Debug mode it now panics.

I'm relying on signed integers to be in two's complement form.

Edit: 
* Updated bench numbers with 5000 random numbers for better statistics
* Added mention of debug mode panics on minimum value

### DIFF
--- a/src/integer.rs
+++ b/src/integer.rs
@@ -360,6 +360,47 @@ macro_rules! impl_integer_for_isize {
                 assert_eq!((-6 as $T).gcd(&3), 3 as $T);
                 assert_eq!((-4 as $T).gcd(&-2), 2 as $T);
             }
+            #[test]
+            fn test_gcd_cmp_with_euclidean() {
+                fn euclidean_gcd(mut m: $T, mut n: $T) -> $T {
+                    while m != 0 {
+                        ::std::mem::swap(&mut m, &mut n);
+                        m %= n;
+                    }
+
+                    n.abs()
+                }
+
+                // gcd(-128, b) = 128 is not representable as positive value
+                // for i8
+                for i in -127..127 {
+                    for j in -127..127 {
+                        assert_eq!(euclidean_gcd(i,j), i.gcd(&j));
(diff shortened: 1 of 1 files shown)


Emerentius:
Noticed that `.abs()` panics on the minimum value in debug mode, allowing for a better guarantee of only positive return values, at least in debug mode. Added that to my implementation for signed integers.
Also added some tests

PR #114, opened 2015-09-13
Implement is_nan for the complex type

achanda:
Also run the associated test

### DIFF
--- a/src/complex.rs
+++ b/src/complex.rs
@@ -281,6 +281,12 @@ impl<T: Clone + Float> Complex<T> {
         }
         ((one + self).ln() - (one - self).ln()) / two
     }
+
+    /// Checks if the given complex number is NaN
+    #[inline]
+    pub fn is_nan(self) -> bool {
+        self.re.is_nan() || self.im.is_nan()
+    }
 }
 
 macro_rules! forward_val_val_binop {
@@ -514,11 +520,9 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
-    #[ignore]
     fn test_inv_zero() {
-        // FIXME #5736: should this really fail, or just NaN?
-        _0_0i.inv();
+        // FIXME #20: should this really fail, or just NaN?
+        assert!(_0_0i.inv().is_nan());
     }
 
     #[test]
@@ -919,5 +923,20 @@ mod test {
         assert!(::hash(&a) != ::hash(&b));
         assert!(::hash(&b) != ::hash(&c));
         assert!(::hash(&c) != ::hash(&a));
+    }
+
+    #[test]
+    fn test_is_nan() {
+        assert!(!_1_1i.is_nan());
+        let a = Complex::new(f64::NAN, f64::NAN);
+        assert!(a.is_nan());
+    }
+
+    #[test]
+    fn test_is_nan_special_cases() {
+        let a = Complex::new(0f64, f64::NAN);
+        let b = Complex::new(f64::NAN, 0f64);
+        assert!(a.is_nan());
+        assert!(b.is_nan());
     }
 }


hauleth:
What would happen if only one of components is NaN? What would happen if we do `Complex::new(1.0, 2.0) / `Complex::new(0.0, 1.0)`? There should be some tests for that cases too.

Also if you rebase and resolve conflicts.

cuviper:
It uses `arg()`, which is basically `im.atan2(re)`, so a NaN in either should be infectious.  But I agree, it needs more exhaustive testing.

Why `arg()` though?  Why isn't it just `re.is_nan() || im.is_nan()` ?

hauleth:
@cuviper there is also possibility that `atan2` return `NaN` so it should be covered also in some way.

cuviper:
But why is `arg()` (or `atan2()`) of particular interest to `Complex::is_nan()`?  And not `norm()`, for instance?  (I'm probably just missing some mathematical context here.)

hauleth:
There is link in documentation: http://mathworld.wolfram.com/ComplexInfinity.html. I think that is the reason behind that.

cuviper:
Hmm, but infinity is a different concept than NaN.  We could definitely implement `is_infinite()` and `is_finite()` base on that definition though.  It seems to me that `is_nan` should be a simple `||`, and `is_normal` could be written that way too -- these are both not really mathematical concepts, just floating point details.

hauleth:
Agree. This should be just `re.is_nan() || im.is_nan()`.

cuviper:
I'm even having second thoughts about using that wolfram definition of ComplexInfinity.  That's a particular constant "whose complex argument is unknown or undefined".  But this doesn't actually imply the reverse, that this the only thing that's possibly infinite.  For instance, `(1 + inf i).arg()` and `(inf + 1 i).arg()` are both well-defined, `pi/2` and `0` respectively.  Even `(inf + inf i).arg()` gives `pi/4`, although that's pretty suspicious to me.  But `norm()` is infinite in all those cases.

I suggest this simplified set:
* `Complex::is_nan() -> re.is_nan() || im.is_nan()`
* `Complex::is_infinite() -> !is_nan() && (re.is_infinite() || im.is_infinite())`
* `Complex::is_finite() -> re.is_finite() && im.is_finite()`
* `Complex::is_normal() -> re.is_normal() && im.is_normal()`
* `Complex::classify() -> match (re.classify(), im.classify()) { /* combos as above */ }`

achanda:
Thanks all. I will push the nan related change here and the others in another commit.

cuviper:
For classify, `(Zero, Normal)` and vice versa is tough to decide.  Maybe just call those `Normal` too?  But that's a bit of a stretch...

Anyway, NaN is the easiest. :smile:

src/complex.rs
```diff
@@ -514,11 +520,9 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
-    #[ignore]
     fn test_inv_zero() {
         // FIXME #5736: should this really fail, or just NaN?
```
cuviper: While we're here, that's referencing a rust-lang/rust issue; can you please update this to #20?

src/complex.rs
```diff
@@ -920,4 +924,11 @@ mod test {
         assert!(::hash(&b) != ::hash(&c));
         assert!(::hash(&c) != ::hash(&a));
     }
+
+    #[test]
+    fn test_is_nan() {
+        assert!(!(_1_1i.is_nan()));
```
hauleth: No need for extra parens.

src/complex.rs
```diff
@@ -920,4 +924,11 @@ mod test {
...
+        let a = Complex::new(f64::NAN, f64::NAN);
+        assert!(a.is_nan());
+    }
```
hauleth: Add test cases for `(NaN, _)` and `(_, NaN)` and IMHO will be ready to go.

Oh, and split it into several tests. `_1_1i.is_nan()` should be IMHO in different test case.

achanda:
I can use some suggestion on test function names :)

hauleth:
@homu r+

Issue #120, opened 2015-10-04
Add a method on BigInt to get the sign, without also getting bytes


alex:

wrieger93:
You could also use the `signum` method http://doc.rust-lang.org/num/num/traits/trait.Signed.html#method.signum

cuviper:
Closed by #121.

PR #121, opened 2015-10-06
Add a function that returns the sign of a BigInt.

wrieger93:
In reference to #120.


hauleth:
Could you add some tests to it?

cuviper:
It does have doctests, and you can see that in the Travis CI run: `test bigint::sign_0 ... ok`

cuviper:
Merged, thanks!

PR #123, opened 2015-10-16
Check for zero denominator when parsing Ratio

### LINKED ISSUE #122: Parsing "1/0" as Ratio panics
It seems to me it should return an error. `ParseRatioError` would need to be changed into an enum. I can make a PR if this change is acceptable.

murarth:
Closes #122


murarth:
Also, can I request a version bump after this is merged?

cuviper:
This looks good on paper, but I'd like to run it through travis CI.  We still need to fix all that up since the move to the rust-num organization.

cuviper:
I'm going to close and reopen to try to trigger travis CI...

cuviper:
The failure is a known problem with num-macros, not your fault.  Looks good, thanks!

PR #124, opened 2015-10-19
Remove extra parentheses to fix a warning

dripton:
Just fixing a compiler warning.  And making a similar fix inside example code.


cuviper:
Thanks!

PR #128, opened 2015-11-05
Implement some more functions for the complex type

achanda:


achanda:
Something is very wrong about homu

hauleth:
@achanda why you say so?

src/complex.rs
```diff
@@ -287,6 +287,24 @@ impl<T: Clone + Float> Complex<T> {
     pub fn is_nan(self) -> bool {
         self.re.is_nan() || self.im.is_nan()
     }
+
+    /// Checks if the given complex number is infinite
+    #[inline]
+    pub fn is_infinite(self) -> bool {
+        !self.is_nan() && (self.re.is_infinite() || self.im.is_infinite())
```
hauleth: According to [Wolfram MathWorld about complex infinity](http://mathworld.wolfram.com/ComplexInfinity.html) this should be rather `self.arg().is_nan()`.

cuviper: As i said in the other PR, I don't think Wolfram's [`ComplexInfinity`](http://mathworld.wolfram.com/ComplexInfinity.html) is a useful definition for a general `is_infinite`. Yes, that's one concept of infinity, but they also have [`DirectedInfinity`](http://mathworld.wolfram.com/DirectedInfinity.html).  In polar form these are roughly "r e^i(undef)" and "∞ e^iθ" respectively.  We could also look at Maple's [`complex`](http://www.maplesoft.com/support/help/maple/view.aspx?path=complex) and [`cx_infinity`](http://www.maplesoft.com/support/help/maple/view.aspx?path=type%2finfinity) test, which requires both real and imaginary parts to be infinite.

But those are both high-level mathematical languages, not dealing with floating point properties.  For a more pragmatic programming view, we can look at Python's [`cmath`](https://docs.python.org/3/library/cmath.html) `isinf()`, which returns "True if either the real or the imaginary part of x is an infinity, and False otherwise."  The source is [here](https://github.com/python/cpython/blob/master/Modules/cmathmodule.c#L1111); it really is that simple.  I also looked for this function in standard C, C++, Haskell, and Ruby complex types, but didn't find it at all.

achanda:
@hauleth homu has been waiting for status from travis since last few days. I took these snippets from the discussions in the earlier PR.

hauleth:
Yeah, because no one approved this PR yet.  When that happen Homu will start checks. 

--
Łukasz Jan Niemier

Dnia 8 lis 2015 o godz. 05:29 Abhishek Chanda <notifications@github.com> napisał(a):

achanda:
@hauleth @cuviper please let me know if this needs any more work

cuviper:
@homu r+

PR #129, opened 2015-11-08
Add accessor methods to BigInt and BigUint

murarth:


cuviper:
Why do you want this?  I don't want to expose internals without good justification.

murarth:
I'm writing a custom `to_str_radix` function that inserts arbitrary grouping. I had previously been using the public division methods, but that turned out to be noticeably slow, either because it was allocating a new `BigUint` for each digit or because it was dividing `BigUint / BigUint` when I only need `BigUint / u32`.

It doesn't seem like much of a stretch to me, given that the accessors are analogous to existing constructor functions.

murarth:
On second thought, I may not need the frills of the custom `to_str_radix`. But it did turn out to be significantly faster than the current implementation. Perhaps I should merge that instead of this.

cuviper:
I see your point relating to the constructors.  Honestly, I don't really like those either, as I think `BigDigit` should be a private implementation detail, but that cat is out of the bag.  I wonder how often they are actually used -- could we kill that in a new major version?  Anyway...

By all means, if your new `to_str_radix` is a clear improvement, please share!

PR #131, opened 2015-11-09
Improve `to_str_radix` implementation

murarth:


murarth:
On my machine, the `bigint::to_string` benchmark (which needed to be explicitly named in Cargo.toml to be recognized) showed an improvement from ~3,500 ns/iter to ~900 ns/iter.

src/bigint.rs
```diff
@@ -809,100 +809,68 @@ impl_to_biguint!(u16,  FromPrimitive::from_u16);
 impl_to_biguint!(u32,  FromPrimitive::from_u32);
 impl_to_biguint!(u64,  FromPrimitive::from_u64);
 
-// Cribbed from core/fmt/num.rs
-#[derive(Copy, Clone)]
-pub struct RadixFmt {
```
cuviper: Strange that this was `pub`, but I don't see any way for users to create or receive one.  I guess someone *could* have written functions taking `RadixFmt` as parameters, but they would be useless (can only `Clone`/`Copy`/`Display`) and have no callers.  This is still an API break if we're being ultra-cautious, but I think it's ok.

murarth: That is strange. I suppose it was made `pub` by accident. Its only use was in the previous implementation of `to_str_radix`. As you say, it can't be created outside the crate.

cuviper:
@homu r+

PR #133, opened 2015-11-16
[RFC] Some performance improvements

koverstreet:
I added add and subtract methods that take an operand by value, instead of by reference, so they can work in place instead of always allocating new BigInts for the result.

I'd like to hear if anyone knows a better way of doing this, though - with the various wrappers/forwardings doing this for all the operations is going to be a mess and add a ton of boilerplate. I haven't even tried to do a complete job of it yet.

I'm starting to think the sanest thing to do would be to only have in place implements of most of the operations - and have the wrappers clone() if necessary. That way, the only extra work we're doing is a memcpy(), which is a hell of a lot faster than a heap allocation.

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -2,10 +2,33 @@
 
 extern crate test;
 extern crate num;
+extern crate rand;
 
 use std::mem::replace;
 use test::Bencher;
 use num::{BigUint, Zero, One, FromPrimitive};
+use num::bigint::RandBigInt;
+use rand::{SeedableRng, StdRng};
+
+fn multiply_bench(b: &mut Bencher, xbits: usize, ybits: usize) {
+    let seed: &[_] = &[1, 2, 3, 4];
+    let mut rng: StdRng = SeedableRng::from_seed(seed);
+
+    let x = rng.gen_bigint(xbits);
+    let y = rng.gen_bigint(ybits);
+
+    b.iter(|| &x * &y);
+}
+
+fn divide_bench(b: &mut Bencher, xbits: usize, ybits: usize) {
+    let seed: &[_] = &[1, 2, 3, 4];
+    let mut rng: StdRng = SeedableRng::from_seed(seed);
+
+    let x = rng.gen_bigint(xbits);
+    let y = rng.gen_bigint(ybits);
+
+    b.iter(|| &x / &y);
+}
 
 fn factorial(n: usize) -> BigUint {
     let mut f: BigUint = One::one();
@@ -24,6 +47,36 @@ fn fib(n: usize) -> BigUint {
         f0 = replace(&mut f1, f2);
     }
     f0
+}
+
+#[bench]
+fn multiply_0(b: &mut Bencher) {
+    multiply_bench(b, 1 << 8, 1 << 8);
+}
+
+#[bench]
+fn multiply_1(b: &mut Bencher) {
+    multiply_bench(b, 1 << 8, 1 << 16);
+}
+
+#[bench]
+fn multiply_2(b: &mut Bencher) {
+    multiply_bench(b, 1 << 16, 1 << 16);
+}
+
+#[bench]
+fn divide_0(b: &mut Bencher) {
+    divide_bench(b, 1 << 8, 1 << 6);
+}
+
+#[bench]
+fn divide_1(b: &mut Bencher) {
+    divide_bench(b, 1 << 12, 1 << 8);
+}
+
+#[bench]
+fn divide_2(b: &mut Bencher) {
+    divide_bench(b, 1 << 16, 1 << 12);
 }
 
 #[bench]
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -116,6 +116,38 @@ pub mod big_digit {
     }
 }
 
+/*
+ * Generic functions for add/subtract/multiply with carry/borrow:
+ */
+
+// Add with carry:
+#[inline]
+fn adc(a: BigDigit, b: BigDigit, carry: &mut BigDigit) -> BigDigit {
+    let (hi, lo) = big_digit::from_doublebigdigit(
+        (a as DoubleBigDigit) +
+        (b as DoubleBigDigit) +
+        (*carry as DoubleBigDigit));
+
+    *carry = hi;
+    lo
+}
+
+// Subtract with borrow:
+#[inline]
+fn sbb(a: BigDigit, b: BigDigit, borrow: &mut BigDigit) -> BigDigit {
+    let (hi, lo) = big_digit::from_doublebigdigit(
+        big_digit::BASE
+        + (a as DoubleBigDigit)
+        - (b as DoubleBigDigit)
+        - (*borrow as DoubleBigDigit));
+    /*
+       hi * (base) + lo == 1*(base) + ai - bi - borrow
+       => ai - bi - borrow < 0 <=> hi == 0
+       */
+    *borrow = if hi == 0 { 1 } else { 0 };
+    lo
+}
+
 /// A big unsigned integer type.
 ///
 /// A `BigUint`-typed value `BigUint { data: vec!(a, b, c) }` represents a number
@@ -467,69 +499,112 @@ impl Unsigned for BigUint {}
 
 forward_all_binop_to_val_ref_commutative!(impl Add for BigUint, add);
 
+// Only for the Add impl:
+#[must_use]
+#[inline]
+fn __add2(a: &mut [BigDigit], b: &[BigDigit]) -> BigDigit {
+    let mut b_iter = b.iter();
+    let mut carry = 0;
+
+    for ai in a.iter_mut() {
+        if let Some(bi) = b_iter.next() {
+            *ai = adc(*ai, *bi, &mut carry);
+        } else if carry != 0 {
+            *ai = adc(*ai, 0, &mut carry);
+        } else {
+            break;
+        }
+    }
+
+    debug_assert!(b_iter.next() == None);
+    carry
+}
+
+/// /Two argument addition of raw slices:
+/// a += b
+///
+/// The caller _must_ ensure that a is big enough to store the result - typically this means
+/// resizing a to max(a.len(), b.len()) + 1, to fit a possible carry.
+fn add2(a: &mut [BigDigit], b: &[BigDigit]) {
+    let carry = __add2(a, b);
+
+    debug_assert!(carry == 0);
+}
+
+/*
+ * We'd really prefer to avoid using add2/sub2 directly as much as possible - since they make the
+ * caller entirely responsible for ensuring a's vector is big enough, and that the result is
+ * normalized, they're rather error prone and verbose:
+ *
+ * We could implement the Add and Sub traits for BigUint + BigDigit slices, like below - this works
+ * great, except that then it becomes the module's public interface, which we probably don't want:
+ *
+ * I'm keeping the code commented out, because I think this is worth revisiting:
+
+impl<'a> Add<&'a [BigDigit]> for BigUint {
+    type Output = BigUint;
+
+    fn add(mut self, other: &[BigDigit]) -> BigUint {
+        if self.data.len() < other.len() {
+            let extra = other.len() - self.data.len();
+            self.data.extend(repeat(0).take(extra));
+        }
+
+        let carry = __add2(&mut self.data[..], other);
+        if carry != 0 {
+            self.data.push(carry);
+        }
+
+        self
+    }
+}
+ */
+
 impl<'a> Add<&'a BigUint> for BigUint {
     type Output = BigUint;
 
-    fn add(self, other: &BigUint) -> BigUint {
-        let mut sum = self.data;
-
-        if other.data.len() > sum.len() {
-            let additional = other.data.len() - sum.len();
-            sum.reserve(additional);
-            sum.extend(repeat(ZERO_BIG_DIGIT).take(additional));
-        }
-        let other_iter = other.data.iter().cloned().chain(repeat(ZERO_BIG_DIGIT));
-
-        let mut carry = 0;
-        for (a, b) in sum.iter_mut().zip(other_iter) {
-            let d = (*a as DoubleBigDigit)
-                + (b as DoubleBigDigit)
-                + (carry as DoubleBigDigit);
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            carry = hi;
-            *a = lo;
+    fn add(mut self, other: &BigUint) -> BigUint {
+        if self.data.len() < other.data.len() {
+            let extra = other.data.len() - self.data.len();
+            self.data.extend(repeat(0).take(extra));
         }
 
-        if carry != 0 { sum.push(carry); }
-        BigUint::new(sum)
+        let carry = __add2(&mut self.data[..], &other.data[..]);
+        if carry != 0 {
+            self.data.push(carry);
+        }
+
+        self
     }
 }
 
 forward_all_binop_to_val_ref!(impl Sub for BigUint, sub);
 
+fn sub2(a: &mut [BigDigit], b: &[BigDigit]) {
+    let mut b_iter = b.iter();
+    let mut borrow = 0;
+
+    for ai in a.iter_mut() {
+        if let Some(bi) = b_iter.next() {
+            *ai = sbb(*ai, *bi, &mut borrow);
+        } else if borrow != 0 {
+            *ai = sbb(*ai, 0, &mut borrow);
+        } else {
+            break;
+        }
+    }
+
+    /* note: we're _required_ to fail on underflow */
+    assert!(borrow == 0 && b_iter.all(|x| *x == 0),
+            "Cannot subtract b from a because b is larger than a.");
+}
+
 impl<'a> Sub<&'a BigUint> for BigUint {
     type Output = BigUint;
 
-    fn sub(self, other: &BigUint) -> BigUint {
-        let mut diff = self.data;
-        let other = &other.data;
-        assert!(diff.len() >= other.len(), "arithmetic operation overflowed");
-
-        let mut borrow: DoubleBigDigit = 0;
-        for (a, &b) in diff.iter_mut().zip(other.iter()) {
-            let d = big_digit::BASE - borrow
-                + (*a as DoubleBigDigit)
-                - (b as DoubleBigDigit);
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            /*
-            hi * (base) + lo == 1*(base) + ai - bi - borrow
-            => ai - bi - borrow < 0 <=> hi == 0
-             */
-            borrow = if hi == 0 { 1 } else { 0 };
-            *a = lo;
-        }
-
-        for a in &mut diff[other.len()..] {
-            if borrow == 0 { break }
-            let d = big_digit::BASE - borrow
-                + (*a as DoubleBigDigit);
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            borrow = if hi == 0 { 1 } else { 0 };
-            *a = lo;
-        }
-
-        assert!(borrow == 0, "arithmetic operation overflowed");
-        BigUint::new(diff)
+    fn sub(mut self, other: &BigUint) -> BigUint {
+        sub2(&mut self.data[..], &other.data[..]);
+        self.normalize()
     }
 }
 
@@ -982,12 +1057,8 @@ impl BigUint {
     ///
     /// The digits are in little-endian base 2^32.
     #[inline]
-    pub fn new(mut digits: Vec<BigDigit>) -> BigUint {
-        // omit trailing zeros
-        while let Some(&0) = digits.last() {
-            digits.pop();
-        }
-        BigUint { data: digits }
+    pub fn new(digits: Vec<BigDigit>) -> BigUint {
+        BigUint { data: digits }.normalize()
     }
 
     /// Creates and initializes a `BigUint`.
@@ -1175,6 +1246,16 @@ impl BigUint {
         if self.is_zero() { return 0; }
         let zeros = self.data.last().unwrap().leading_zeros();
         return self.data.len()*big_digit::BITS - zeros as usize;
+    }
+
+    /// Strips off trailing zero bigdigits - comparisons require the last element in the vector to
+    /// be nonzero.
+    #[inline]
+    fn normalize(mut self) -> BigUint {
+        while let Some(&0) = self.data.last() {
+            self.data.pop();
+        }
+        self
     }
 }
 
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -148,6 +148,16 @@ fn sbb(a: BigDigit, b: BigDigit, borrow: &mut BigDigit) -> BigDigit {
     lo
 }
 
+#[inline]
+fn mac_with_carry(a: BigDigit, b: BigDigit, c: BigDigit, carry: &mut BigDigit) -> BigDigit {
+    let (hi, lo) = big_digit::from_doublebigdigit(
+        (a as DoubleBigDigit) +
+        (b as DoubleBigDigit) * (c as DoubleBigDigit) +
+        (*carry as DoubleBigDigit));
+    *carry = hi;
+    lo
+}
+
 /// A big unsigned integer type.
 ///
 /// A `BigUint`-typed value `BigUint { data: vec!(a, b, c) }` represents a number
@@ -172,18 +182,25 @@ impl PartialOrd for BigUint {
     }
 }
 
+fn cmp_slice(a: &[BigDigit], b: &[BigDigit]) -> Ordering {
+    debug_assert!(a.last() != Some(&0));
+    debug_assert!(b.last() != Some(&0));
+
+    let (a_len, b_len) = (a.len(), b.len());
+    if a_len < b_len { return Less; }
+    if a_len > b_len { return Greater;  }
+
+    for (&ai, &bi) in a.iter().rev().zip(b.iter().rev()) {
+        if ai < bi { return Less; }
+        if ai > bi { return Greater; }
+    }
+    return Equal;
+}
+
 impl Ord for BigUint {
     #[inline]
     fn cmp(&self, other: &BigUint) -> Ordering {
-        let (s_len, o_len) = (self.data.len(), other.data.len());
-        if s_len < o_len { return Less; }
-        if s_len > o_len { return Greater;  }
-
-        for (&self_i, &other_i) in self.data.iter().rev().zip(other.data.iter().rev()) {
-            if self_i < other_i { return Less; }
-            if self_i > other_i { return Greater; }
-        }
-        return Equal;
+        cmp_slice(&self.data[..], &other.data[..])
     }
 }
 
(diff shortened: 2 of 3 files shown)


cuviper:
CI failed on rust 1.0 .. We'll want to keep that working if possible.

I actually have been working on something similar.  I'll try to clean that up and push to a branch soon so we can compare/merge approaches.

Do you have comparative benchmark results to share?

koverstreet:
On Mon, Nov 16, 2015 at 01:50:04AM -0800, Josh Stone wrote:
> CI failed on rust 1.0 .. We'll want to keep that working if possible.

Yeah, I just saw that...

> I actually have been working on something similar.  I'll try to clean that up and push to a branch soon so we can compare/merge approaches.

Figures :) I'm definitely interested in seeing what you did. This is my "screw
around and learn Rust" project.

I'm also toying with inline assembly for the inner loops - mainly for the
purpose of getting away from DoubleBigDigit so all the arithmatic can be done in
the native word size. So far just toying, though.

> Do you have comparative benchmark results to share?

Yeah, in the patch messages.

koverstreet:
Just got it working with inline asm and BigDigit size == usize, if anyone wants to take a look.

https://github.com/koverstreet/num/commits/asm

Very hacky but it does give a nice performance boost on the multiply benchmark.

koverstreet:
Found some new performance improvements and rolled them into the patch. The current numbers are:

Before:
multiply_0: 7,137 ns/iter (+/- 620)
multiply_1: 2,104,565 ns/iter (+/- 255,208)
multiply_2: 51,620,572 ns/iter (+/- 2,707,818)
    
After:
multiply_0: 634 ns/iter (+/- 61)
multiply_1: 60,992 ns/iter (+/- 6,843)
multiply_2: 6,849,681 ns/iter (+/- 388,000)

cuviper:
OK, I pushed my stuff to [cuviper:wip](https://github.com/cuviper/num/tree/wip).  I think I've got the sort of val/ref forwarding changes you were thinking of.  For the most part, my changes are more incremental tweaks than what you have.  My performance improvement to Add/Sub is similar to yours, I think, but I definitely didn't improve Mul as much as you did.  Take a look -- if it's not too hard for you to refactor your improvements on top of mine, that would be awesome... I think my forwarding simplifies it a bit, but let me know what you think.

As for asm stuff, that's a tricky door to open.  It's a feature only for rust nightly, of course, and only for archs you implement, so there still has to be a pure-Rust fallback.  I think if you want to explore this route, you'd be better off working on [ramp](https://github.com/Aatch/ramp).

koverstreet:
On Mon, Nov 16, 2015 at 11:23:50AM -0800, Josh Stone wrote:
> OK, I pushed my stuff to [cuviper:wip](https://github.com/cuviper/num/tree/wip).  I think I've got the sort of val/ref forwarding changes you were thinking of.  For the most part, my changes are more incremental tweaks than what you have.  My performance improvement to Add/Sub is similar to yours, I think, but I definitely didn't improve Mul as much as you did.  Take a look -- if it's not too hard for you to refactor your improvements on top of mine, that would be awesome... I think my forwarding simplifies it a bit, but let me know what you think.

Yeah, I can do that. Any idea when your changes will go in?

> As for asm stuff, that's a tricky door to open.  It's a feature only for rust nightly, of course, and only for archs you implement, so there still has to be a pure-Rust fallback.  I think if you want to explore this route, you'd be better off working on [ramp](https://github.com/Aatch/ramp).

Oh, I'm not in a rush to get the asm stuff in. But I do like my work to be used,
so I'd prefer to just do whatever is necessary to get it into num when it's
ready, no offense to the ramp project :)

Is it possible to use #[cfg] to switch off the rust version? Or would we be
stuck waiting until the oldest version of rust we have to support supports
whatever features we want to use?

cuviper:
There you go, I pushed my changes.  Thanks homu. :)

I don't know any way to `#[cfg]` the rust version -- I wish there were.  We could add a manual `asm` or `nightly` feature in cargo, but nothing automatic that I'm aware of.

hauleth:
@cuviper there is none. Only way to support asm in `num` would be Carfo feature flag.

koverstreet:
On Wed, Nov 18, 2015 at 10:06:03PM -0800, Josh Stone wrote:
> There you go, I pushed my changes.  Thanks homu. :)
> 
> I don't know any way to `#[cfg]` the rust version -- I wish there were.  We could add a manual `asm` or `nightly` feature in cargo, but nothing automatic that I'm aware of.

Damn, that sounds like a major oversight. If I'm feeling particularly ambitious
maybe I'll work on adding that to rustc.

That said - I also wrote a patch that implements all the low level arithmetic
(the same ones the asm patch implements) without using DoubleBigDigit, so
BigDigit can be changed to usize - that actually gets us most of the performance
of the asm patch (which is still rather simple and dumb).

I pushed that patch out, and also rebased everything on top of your stuff.

cuviper:
Let's step back a little and keep focus in this PR, and especially do what we can without breaking API.  We can do more in followup PRs.  So let's keep the basic add/sub/mul optimizations (first four commits).  The ops against primitives can be their own PR, a lot like what #125 did for complex numbers.  The BigDigit changes constitute an API break -- which I am open to, but we'd need to bump semver, and if we do this I'd like to hide BigDigit entirely as an implementation detail.  (But perhaps some of your optimizations avoiding DoubleBigDigit are still usable right now, if they still help with u32 BigDigit.)

I'll add a few more inline comments too...

benches/bigint.rs
```diff
@@ -27,6 +50,36 @@ fn fib(n: usize) -> BigUint {
...
+
+#[bench]
+fn divide_1(b: &mut Bencher) {
+    divide_bench(b, 1 << 12, 1 << 8);
+}
+
+#[bench]
+fn divide_2(b: &mut Bencher) {
+    divide_bench(b, 1 << 16, 1 << 12);
+}
```
cuviper: This last one takes painfully long to benchmark, even after all of your optimizations.  Let's leave it out or `#[ignore]` it for now.  (Then use `--ignored` to run it anyway.)

src/bigint.rs
```diff
@@ -79,43 +79,124 @@ use {Num, Unsigned, CheckedAdd, CheckedSub, CheckedMul, CheckedDiv, Signed, Zero
...
 
 /// A `BigDigit` is a `BigUint`'s composing element.
-pub type BigDigit = u32;
-
-/// A `DoubleBigDigit` is the internal type used to do the computations.  Its
-/// size is the double of the size of `BigDigit`.
-pub type DoubleBigDigit = u64;
-
-pub const ZERO_BIG_DIGIT: BigDigit = 0;
+pub type BigDigit = usize;
```
cuviper: We need to keep all `pub` stuff until we're ready to bump semver.

src/bigint.rs
```diff
@@ -125,6 +206,8 @@ pub struct BigUint {
     data: Vec<BigDigit>
 }
 
+type BigUintS<'a> = &'a [BigDigit];
```
cuviper: I don't think this type alias is very helpful, especially since you're still writing out `&mut [BigDigit]` when it needs to be mutable.  Let's leave it all explicit -- it's only a few more characters to write `&[BigDigit]` anyway.

src/bigint.rs
```diff
@@ -467,146 +546,352 @@ impl Unsigned for BigUint {}
...
+    debug_assert!(carry == 0);
+}
+
 impl<'a> Add<&'a BigUint> for BigUint {
     type Output = BigUint;
 
-    fn add(self, other: &BigUint) -> BigUint {
-        let mut sum = self.data;
+    fn add(mut self, other: &BigUint) -> BigUint {
+        let new_len = cmp::max(self.data.len(), other.data.len()) + 1;
```
cuviper: Hmm, this *always* grows `self.data`, even when the extra digit might not be needed.  It may be that the vector often has that extra capacity anyway, but... can we keep the property that the extra carry is only pushed when we know it really overflowed?  Maybe let `add2` return the leftover carry, for instance.

src/bigint.rs
```diff
@@ -467,146 +546,352 @@ impl Unsigned for BigUint {}
...
-                - (b as DoubleBigDigit);
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            /*
-            hi * (base) + lo == 1*(base) + ai - bi - borrow
-            => ai - bi - borrow < 0 <=> hi == 0
-             */
-            borrow = if hi == 0 { 1 } else { 0 };
-            *a = lo;
-        }
+forward_all_binop_to_ref_ref!(impl Mul for BigUint, mul);
```
cuviper: Just related to this, since you reverted back to simple ref-ref, we could revert BigInt's Mul back to simple forwarding too, rather than the `bigint_mul` dance I had added.

koverstreet:
Thanks for the review feedback.

On Sun, Nov 22, 2015 at 05:31:23PM -0800, Josh Stone wrote:
> Let's step back a little and keep focus in this PR, and especially do what we can without breaking API.  We can do more in followup PRs.  So let's keep the basic add/sub/mul optimizations (first four commits).  The ops against primitives can be their own PR, a lot like what #125 did for complex numbers. 

Alright, sounds good.

I do have a patch to improve divide performance - do you mind including that, to
fix the issue with the benchmark taking too long to run?

The only thing I haven't changed per your comments is add/add2 and the way carry
is handled - it's a relatively minor issue, but I feel better about handling the
carry within add2 instead of returning it.

Reason being in the other places where add2 is used (multiply/divide), if they
don't size a correctly to hold the result then that's a serious bug (we'd return
an incorrect result) - and the way I've got it now we always catch that with an
assert, but if add2 is returning a carry bit then it's up to the caller to
check.

We could handle that with more wrapper functions, but... it just seems not worth
it as long as the extra push/pop isn't an actual performance issue (and it
really shouldn't be, having to realloc the vec won't be the common case).

So I wouldn't mind if it's changed later but I lean towards leaving it with the
extra push/pop for now.

Aside from that - I incorporated all your review feedback, and pushed with just
the performance optimizations. Let me know if that works, or if you want me to
split the division optimizations off (but I hate having stuff multiply into a
crapton of different pull requests).

> The BigDigit changes constitute an API break -- which I am open to, but we'd need to bump semver, and if we do this I'd like to hide BigDigit entirely as an implementation detail.  (But perhaps some of your optimizations avoiding DoubleBigDigit are still usable right now, if they still help with u32 BigDigit.)

*nod* Not used to having to consider API breakages.

BTW, what I'd really like is to come up with a good way of separating out the
implementation of the low level arithmetic (into another mod?) with a nice
abstract interface, in a way that's easy to select which one to build with.

And then ideally keep the DoubleBigDigit implementation, along with my BigDigit
= usize implementation and hopefully eventually fancy asm ones. Partly because I
can completely believe the DoubleBigDigit implementation could be faster on
different cpus, but also because it's definitely more readable.

So maybe something to consider if/when we do an API break, if you know a nice
clean way of doing all that.

cuviper:
"The benchmark is too slow?  Let me improve it by 500x..."  That's just scary.  I feel I now need to pick over that with a fine-toothed comb, but who am I to question Knuth's methods.  :smile:

Regarding the carry push/pop, it's a small point, but I'm afraid I still disagree.  I think we'd be better served by having two variants of `add2` -- one that returns the extra carry, and one that asserts there is no carry (probably built on the former).  If you don't want to do this, I probably will, unless you can make a better case why this is harmful.  I expect it should perform no worse -- occasionally even better for avoiding capacity growth -- and it's conceptually cleaner to me.

Regarding API and implementations -- technically we could do whatever we want in implementation even now, as long as the public interfaces remain.  `BigDigit` and `DoubleBigDigit` could become vestigial types with no bearing on the true innards anymore.  However, `new(digits: Vec<BigDigit>)` does have a strong implication about what it's doing, that this should be a fast constructor by reusing that vector, so we probably don't want to stray too far.  When we finally do an API break, we should hide these innards completely so we're free to implement however we see best, rather than trying to juggle multiple implementations as you seem to suggest.

Patch comments inbound.  (Some of these may not even be for new changes, just things I didn't notice before.)

benches/bigint.rs
```diff
@@ -2,10 +2,33 @@
...
+
+fn divide_bench(b: &mut Bencher, xbits: usize, ybits: usize) {
+    let seed: &[_] = &[1, 2, 3, 4];
+    let mut rng: StdRng = SeedableRng::from_seed(seed);
+
+    let x = rng.gen_bigint(xbits);
+    let y = rng.gen_bigint(ybits);
+
+    b.iter(|| &x / &y);
+}
```
cuviper: How about some pre-iter assertions that these huge divisions are actually correct?  e.g. check the round trip to multiply and add the remainder back.

benches/bigint.rs
cuviper: The new performance is impressive enough that we can keep this after all.

src/bigint.rs
```diff
@@ -116,6 +116,54 @@ pub mod big_digit {
...
+    *carry = hi;
+    lo
+}
+
+#[inline]
+fn div_wide(hi: BigDigit, lo: BigDigit, divisor: BigDigit) -> (BigDigit, BigDigit) {
+    let lhs = big_digit::to_doublebigdigit(hi, lo);
+    let rhs = divisor as DoubleBigDigit;
+    ((lhs / rhs) as BigDigit, (lhs % rhs) as BigDigit)
+}
```
cuviper: Considered alone, this looks like a bad assumption that `lhs/rhs` will always fit into `BigDigit`.  But since this is only called from `div_rem_digit` with prior remainders in `hi`, it should be fine.  Perhaps this deserves a comment and `debug_assert!(hi < divisor)`, or just fold this code into `div_rem_digit` where it's known to be safe.

src/bigint.rs
```diff
@@ -140,18 +188,25 @@ impl PartialOrd for BigUint {
...
+    debug_assert!(b.last() != Some(&0));
+
+    let (s_len, o_len) = (a.len(), b.len());
+    if s_len < o_len { return Less; }
+    if s_len > o_len { return Greater;  }
+
+    for (&self_i, &other_i) in a.iter().rev().zip(b.iter().rev()) {
+        if self_i < other_i { return Less; }
+        if self_i > other_i { return Greater; }
+    }
```
cuviper: Let's give these a/b-names instead of s/o/self/other from the original code.

src/bigint.rs
```diff
@@ -467,146 +522,303 @@ impl Unsigned for BigUint {}
...
+            *ai = adc(*ai, 0, &mut carry);
+        } else {
+            break;
+        }
+    }
+
+    assert!(b_iter.next() == None && carry == 0);
+}
+
+impl<'a> Add<&'a [BigDigit]> for BigUint {
```
cuviper: This new impl is visible in the public API, which is undesirable.  AFAICS this only used by the normal Add implementation, so let's keep it as one.

src/bigint.rs
```diff
@@ -467,146 +522,303 @@ impl Unsigned for BigUint {}
...
-        for (a, b) in sum.iter_mut().zip(other_iter) {
-            let d = (*a as DoubleBigDigit)
-                + (b as DoubleBigDigit)
-                + (carry as DoubleBigDigit);
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            carry = hi;
-            *a = lo;
+        while self.data.len() < new_len {
+            self.data.push(0);
         }
```
cuviper: The advantage of `.extend(repeat(0).take(foo))` is that it calls `reserve` up front before pushing.  (The old code was redundant in that.)  You've used repeat.take elsewhere, why not here?

src/bigint.rs
```diff
@@ -467,146 +522,303 @@ impl Unsigned for BigUint {}
...
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            borrow = if hi == 0 { 1 } else { 0 };
-            *a = lo;
-        }
+    /* note: we're _required_ to fail on underflow */
+    assert!(borrow == 0 && b_iter.all(|x| *x == 0),
+            "Cannot subtract b from a because b is larger than a.");
+}
+
+impl<'a> Sub<&'a [BigDigit]> for BigUint {
```
cuviper: Ditto on undesirable public API with slices -- but I see this one is used both in the regular `Sub` and in `sub_sign`.  Since the only thing it does over than `sub2` is normalize, let's just do that in the callers.  Or else make a new private function instead of this new `impl Sub`.

src/bigint.rs
```diff
@@ -467,146 +522,303 @@ impl Unsigned for BigUint {}
...
-            let (s1, n1) = sub_sign(s_hi, s_lo);
-            let (s2, n2) = sub_sign(o_hi, o_lo);
-            match (s1, s2) {
-                (Equal, _) | (_, Equal) => &hh + &ll,
-                (Less, Greater) | (Greater, Less) => &hh + &ll + (n1 * n2),
-                (Less, Less) | (Greater, Greater) => &hh + &ll - (n1 * n2)
-            }
-        };
+            i != 0
+        } { i = i - 1; }
```
cuviper: Weird, I didn't even know you could write a while-loop this way...

Anyway, it seems wasteful not to use `cmp_slice`.  A quick experiment to just normalize the zeros away first seems to work just as fast:

```
    let a_len = a.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1);
    let b_len = b.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1);
    match cmp_slice(&a[..a_len], &b[..b_len]) { ... }
```

src/bigint.rs
```diff
@@ -467,146 +522,303 @@ impl Unsigned for BigUint {}
...
+            },
+            Minus   => {
+                mac3(&mut acc[b..], &j0.data.data[..], &j1.data.data[..]);
+            },
+            NoSign  => (),
+        }
+    }
+}
+
+fn mul3(x: &[BigDigit], y: &[BigDigit]) -> BigUint {
```
cuviper: Just curious, what is the 3 in `mul3`?  But anyway, this function is only used by `Mul::mul`, which itself isn't doing anything else, which raises the question why this should be separate at all.

src/bigint.rs
```diff
@@ -709,77 +921,88 @@ impl Integer for BigUint {
...
+             *     q += q0
+             *     a -= q0 * b
+             *
+             * and then iterate until a < b. Then, (q, a) will be our desired quotient and
+             * remainder.
+             *
+             * q0, our guess, is calculated by dividing the last few digits of a by the last digit
+             * of b - this should give us a guess that is "close" to the actual quotient, but is
+             * possibly greater than the actual quotient. If q0 * b > a, we simply use iterated
+             * subtraction until we have a guess such that q8 & b <= a.
```
cuviper: typo, s/q8/q0/

src/bigint.rs
```diff
@@ -709,77 +921,88 @@ impl Integer for BigUint {
...
+             */
+
+            let bn = *b.data.last().unwrap();
+            let q_len = a.data.len() - b.data.len() + 1;
+            let mut q: BigUint = BigUint { data: Vec::with_capacity(q_len) };
+
+            q.data.extend(repeat(0).take(q_len));
+
+            /* We reuse the same temporary to avoid hitting the allocator in our inner loop: */
+            let mut tmp: BigUint = BigUint { data: Vec::with_capacity(2) };
```
cuviper: Why 2?  Is this the worst case needed, or can that be guessed?

src/bigint.rs
```diff
@@ -709,77 +921,88 @@ impl Integer for BigUint {
...
-                // d += d0;
-                d = d + d0;
-                // FIXME(#5992): assignment operator overloads
-                // m -= prod;
-                m = m - prod;
+
+                /* just avoiding a heap allocation: */
+                let mut a0 = tmp;
+                a0.data.truncate(0);
+                a0.data.extend(&a.data[offset..]);
```
cuviper: This failed CI on Rust 1.0.  Later Rust has for convenience:

    impl<'a, T> Extend<&'a T> for Vec<T> where T: Copy + 'a

But for maintaining 1.0 compatibility, we'll have to explicitly call `a.data[offset..].iter().cloned()`.  This shouldn't make any difference in performance, just syntactic sugar.

koverstreet:
On Mon, Nov 30, 2015 at 03:38:14PM -0800, Josh Stone wrote:
> "The benchmark is too slow?  Let me improve it by 500x..."  That's just scary.
> I feel I now need to pick over that with a fine-toothed comb, but who am I to
> question Knuth's methods.  :smile:

:D

The code isn't fresh in my head anymore but I had a suspicion as I was working
on it that the old code had the algorithm slightly wrong - at least, the logic
didn't quite match Knuth's algorithm but I didn't bother to try and figure out
if it affected the behaviour. I dug out my copy of TAOCP and just stuck to
Knuth's algorithm when I was working on it.

> Regarding the carry push/pop, it's a small point, but I'm afraid I still disagree.  I think we'd be better served by having two variants of `add2` -- one that returns the extra carry, and one that asserts there is no carry (probably built on the former).  If you don't want to do this, I probably will, unless you can make a better case why this is harmful.  I expect it should perform no worse -- occasionally even better for avoiding capacity growth -- and it's conceptually cleaner to me.

Alright, I can do the wrapper thing.

> Regarding API and implementations -- technically we could do whatever we want in implementation even now, as long as the public interfaces remain.  `BigDigit` and `DoubleBigDigit` could become vestigial types with no bearing on the true innards anymore.  However, `new(digits: Vec<BigDigit>)` does have a strong implication about what it's doing, that this should be a fast constructor by reusing that vector, so we probably don't want to stray too far.  When we finally do an API break, we should hide these innards completely so we're free to implement however we see best, rather than trying to juggle multiple implementations as you seem to suggest.

Good point. I went and split out the implementations of the low level arithmetic
into different modules like I was talking about:

https://github.com/koverstreet/num/commit/ff06399293e90ffc21178f307aaa11546e6b85ff

btw - the operations I have the low level arithmatic module exposing right now
(like div_wide, which I saw you mentioned in another email) are ones that match
what you can do with (x86 at least, I don't know arm) assembly - that is,
there's no good reasons these operations available as intrinsics and my hope is
that's what happens eventually.

Although for add with carry/subtract with borrow, my language nerd friend (who's
working on Swift, and thus gets to interact with the llvm guys) tells me that
llvm can't really model instructions with two different outputs - the carry bit
- so that may never happen.

Re: the vector constructor, if there's actual outside demand for that kind of
functionality (and I suppose there probably is) then it probably makes sense to
support it for multiple sizes of digit (say someone's bridging between gmp or
some other library and num; they're going to be stuck with whatever size digit
that other library uses).

We could do fast in place converion of vectors of arbitrary size digits to
biguints with a tiny bit of unsafe code - just cast (is it transmute in rust?)
the vector to the new type and then just fix the endianness thing in place.

So if you want to keep that functionality and hide BigDigit I could do that.

benches/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:38:38PM -0800, Josh Stone wrote:
> > +
> > +    let x = rng.gen_bigint(xbits);
> > +    let y = rng.gen_bigint(ybits);
> > +
> > +    b.iter(|| &x * &y);
> > +}
> > +
> > +fn divide_bench(b: &mut Bencher, xbits: usize, ybits: usize) {
> > +    let seed: &[_] = &[1, 2, 3, 4];
> > +    let mut rng: StdRng = SeedableRng::from_seed(seed);
> > +
> > +    let x = rng.gen_bigint(xbits);
> > +    let y = rng.gen_bigint(ybits);
> > +
> > +    b.iter(|| &x / &y);
> > +}
> 
> How about some pre-iter assertions that these huge divisions are actually correct?  e.g. check the round trip to multiply and add the remainder back.

Well, then it wouln't be just benchmarking division - I do have a test that does
that though, I didn't push it because it takes way too long to run if you don't
do a release build.

I don't know a good way of putting in these kinds of torture tests where you
want it to run "as long as you have time for", not any fixed amount of time. I
guess I could just make it a #[ignore] test, like you suggested for the other
division test.

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:39:08PM -0800, Josh Stone wrote:
> > +#[inline]
> > +fn mac_with_carry(a: BigDigit, b: BigDigit, c: BigDigit, carry: &mut BigDigit) -> BigDigit {
> > +    let (hi, lo) = big_digit::from_doublebigdigit(
> > +        (a as DoubleBigDigit) +
> > +        (b as DoubleBigDigit) * (c as DoubleBigDigit) +
> > +        (*carry as DoubleBigDigit));
> > +    *carry = hi;
> > +    lo
> > +}
> > +
> > +#[inline]
> > +fn div_wide(hi: BigDigit, lo: BigDigit, divisor: BigDigit) -> (BigDigit, BigDigit) {
> > +    let lhs = big_digit::to_doublebigdigit(hi, lo);
> > +    let rhs = divisor as DoubleBigDigit;
> > +    ((lhs / rhs) as BigDigit, (lhs % rhs) as BigDigit)
> > +}
> 
> Considered alone, this looks like a bad assumption that `lhs/rhs` will always fit into `BigDigit`.  But since this is only called from `div_rem_digit` with prior remainders in `hi`, it should be fine.  Perhaps this deserves a comment and `debug_assert!(hi < divisor)`, or just fold this code into `div_rem_digit` where it's known to be safe.

Like I mentioned, this is implementing exactly what the x86 div instruction does
- I'll add a comment explaining the caller has to know the result will fit into
a BigDigit and that that's not true in general.

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:39:34PM -0800, Josh Stone wrote:
> > +    let mut carry = 0;
> > +
> > +    for ai in a.iter_mut() {
> > +        if let Some(bi) = b_iter.next() {
> > +            *ai = adc(*ai, *bi, &mut carry);
> > +        } else if carry != 0 {
> > +            *ai = adc(*ai, 0, &mut carry);
> > +        } else {
> > +            break;
> > +        }
> > +    }
> > +
> > +    assert!(b_iter.next() == None && carry == 0);
> > +}
> > +
> > +impl<'a> Add<&'a [BigDigit]> for BigUint {
> 
> This new impl is visible in the public API, which is undesirable.  AFAICS this only used by the normal Add implementation, so let's keep it as one.

Argh, I didn't think about that. Is it possible to keep it private? I know it's
not used in this patch series, but I added it because I'm really trying to get
away from using add2/sub2 directly - due to a rather embarassing bug I found the
other night (in sub_sign), which never would've slipped by except for all the
boilerplate from using sub2 directly.

kind of an aside - btw, do you work on the compiler at all? the other usage of
add2() that I haven't been able to get away from is all being used to optimize
code like

acc += b << shift

If we a shl_accumulate method, and some way to teach the compiler/optimizer to
replace (maybe think of it as a type of specialization?) expressons like that
with our shl_accumulate method - we could write a lot of code more naturally,
and the users of this library would benefit too.

You could think of it as generalizing peephole optimizations, I think. Once
you're in SSA form, everything is of the form

f(g(x, y), z)

hand waving a bit and writing it differently than usual - but every value is a
pointer to the thing that produces it, right? So just pretending every value is
a function call.

So I think this might end up being fairly simple: the idea is, if I implement
functions f and g, I can provide a function f_g (or whatever) and tell the
compiler that expressions of the form f(g(x, y), z) can be replaced with calls
to f_g(x, y, z) - which in this case is this shl_accumulate method.

Make any sense? I mean - the compiler could do all these optimizations if the
code was working on primitive types, but because we're doing math with our
special fancy types none of those optimizations are available. It'd be really
cool to be able to write all this math code in a sane natural way.

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:39:45PM -0800, Josh Stone wrote:
> >  
> > -        let mut carry = 0;
> > -        for (a, b) in sum.iter_mut().zip(other_iter) {
> > -            let d = (*a as DoubleBigDigit)
> > -                + (b as DoubleBigDigit)
> > -                + (carry as DoubleBigDigit);
> > -            let (hi, lo) = big_digit::from_doublebigdigit(d);
> > -            carry = hi;
> > -            *a = lo;
> > +        while self.data.len() < new_len {
> > +            self.data.push(0);
> >          }
> 
> The advantage of `.extend(repeat(0).take(foo))` is that it calls `reserve` up front before pushing.  (The old code was redundant in that.)  You've used repeat.take elsewhere, why not here?

No good reason, I switched it to .extend()

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:40:51PM -0800, Josh Stone wrote:
> > -        let (s_hi, s_lo) = cut_at(self,  half_len);
> > -        let (o_hi, o_lo) = cut_at(other.clone(), half_len);
> > -
> > -        let ll = &s_lo * &o_lo;
> > -        let hh = &s_hi * &o_hi;
> > -        let mm = {
> > -            let (s1, n1) = sub_sign(s_hi, s_lo);
> > -            let (s2, n2) = sub_sign(o_hi, o_lo);
> > -            match (s1, s2) {
> > -                (Equal, _) | (_, Equal) => &hh + &ll,
> > -                (Less, Greater) | (Greater, Less) => &hh + &ll + (n1 * n2),
> > -                (Less, Less) | (Greater, Greater) => &hh + &ll - (n1 * n2)
> > -            }
> > -        };
> > +            i != 0
> > +        } { i = i - 1; }
> 
> Weird, I didn't even know you could write a while-loop this way...
> 
> Anyway, it seems wasteful not to use `cmp_slice`.  A quick experiment to just normalize the zeros away first seems to work just as fast:
> 
> ```
>     let a_len = a.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1);
>     let b_len = b.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1);
>     match cmp_slice(&a[..a_len], &b[..b_len]) { ... }
> ```

Oh, I like that.

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:42:43PM -0800, Josh Stone wrote:
> > +             * remainder.
> > +             *
> > +             * q0, our guess, is calculated by dividing the last few digits of a by the last digit
> > +             * of b - this should give us a guess that is "close" to the actual quotient, but is
> > +             * possibly greater than the actual quotient. If q0 * b > a, we simply use iterated
> > +             * subtraction until we have a guess such that q8 & b <= a.
> > +             */
> > +
> > +            let bn = *b.data.last().unwrap();
> > +            let q_len = a.data.len() - b.data.len() + 1;
> > +            let mut q: BigUint = BigUint { data: Vec::with_capacity(q_len) };
> > +
> > +            q.data.extend(repeat(0).take(q_len));
> > +
> > +            /* We reuse the same temporary to avoid hitting the allocator in our inner loop: */
> > +            let mut tmp: BigUint = BigUint { data: Vec::with_capacity(2) };
> 
> Why 2?  Is this the worst case needed, or can that be guessed?

It's the common case, not the worst case: so, we're building up the quotient one
digit at a time, high to low - and as we add a digit q0 to the quotient we
subtract q0 * b from a.

So typically a becomes one digit shorter for every iteration of the main loop;
and as the offset into a we use for calculating our guess decreases in lockstep
with our position in the quotient, there will usually be only two digits in a
past that offset - the digits that get used for a0.

why 2 digits, and not 1? Well, working backwards - we normalized b so that bn
(the number we divide a0 by to generate our guess) was as big as it could be and
fit in a digit - so if a0 was a single digit, it'll usually be smaller than bn
and a0 / bn = (0, a0) - our guess for the quotient will be 0 - and when a digit
in the quotient is 0, a doesn't decrease in that iteration of the loop, and a0
will be one digit longer on the next iteration.

(question for myself: the algorithm assumes/requires that q0 - the guess for a
particular digit of the quotient - is not smaller than the actual quotient. but
when generating that guess, we used a0 (which is < than a) and bn (which is <
than b) - so why is q0 >= q at that digit? brb, checking knuth)

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:39:22PM -0800, Josh Stone wrote:
> > @@ -140,18 +188,25 @@ impl PartialOrd for BigUint {
> >      }
> >  }
> >  
> > +fn cmp_slice(a: &[BigDigit], b: &[BigDigit]) -> Ordering {
> > +    debug_assert!(a.last() != Some(&0));
> > +    debug_assert!(b.last() != Some(&0));
> > +
> > +    let (s_len, o_len) = (a.len(), b.len());
> > +    if s_len < o_len { return Less; }
> > +    if s_len > o_len { return Greater;  }
> > +
> > +    for (&self_i, &other_i) in a.iter().rev().zip(b.iter().rev()) {
> > +        if self_i < other_i { return Less; }
> > +        if self_i > other_i { return Greater; }
> > +    }
> 
> Let's give these a/b-names instead of s/o/self/other from the original code.

Whoops, missed that - fixed

src/bigint.rs
koverstreet: On Mon, Nov 30, 2015 at 03:42:11PM -0800, Josh Stone wrote:
> > +                p.data.extend(repeat(0).take(len));
> > +
> > +                mac3(&mut p.data[..], &j0.data.data[..], &j1.data.data[..]);
> > +                p = p.normalize();
> > +
> > +                sub2(&mut acc[b..], &p.data[..]);
> > +            },
> > +            Minus   => {
> > +                mac3(&mut acc[b..], &j0.data.data[..], &j1.data.data[..]);
> > +            },
> > +            NoSign  => (),
> > +        }
> > +    }
> > +}
> > +
> > +fn mul3(x: &[BigDigit], y: &[BigDigit]) -> BigUint {
> 
> Just curious, what is the 3 in `mul3`?  But anyway, this function is only used by `Mul::mul`, which itself isn't doing anything else, which raises the question why this should be separate at all.

Back when I still had an add3 and needed to distinguish, I was just using the
same convention you use in assembly for talking about ops that overwrite one of
the arguments for the result, or have another argument for the destination (e.g.
two argument add/mul that overwrites an argument for the dest, or 3 argument
that doesn't - or mac3 vs. mac4 (multiply accumulate)). It probably doesn't make
much sense anymore but it still aesthetically pleases me somehow :)

Though if we do ever (hopefully) get alloca() or some other way of moving heap
allocations to the stack, it could make sense to add a mul2() - avoid doing any
heap allocations by putting all the temporaries (including the result) on the
stack, and then return the result by copying over one of the (heap allocated)
arguments.

It does also get used in other patches I have (e.g. divide and conquer division
is able to make use of multiplying slices).

benches/bigint.rs
cuviper: By "pre-iter" I meant before the `b.iter()` call which loops itself.  Just like the `gen_bigint` calls are not included in the benchmark results -- this is just setup, and I'm only suggesting a quick sanity check before hammering at it. :)

src/bigint.rs
cuviper: Right.  And actually x86 div will give you a `#DE` exception if the quotient is too large, which is why I think a `debug_assert!(hi < divisor)` is a good idea here.  I believe that's both the sufficient and necessary condition needed to avoid overflow -- think of long division by hand.

src/bigint.rs
```diff
@@ -140,18 +194,25 @@ impl PartialOrd for BigUint {
...
 }
 
+fn cmp_slice(a: &[BigDigit], b: &[BigDigit]) -> Ordering {
+    debug_assert!(a.last() != Some(&0));
+    debug_assert!(b.last() != Some(&0));
+
+    let (s_len, o_len) = (a.len(), b.len());
+    if s_len < o_len { return Less; }
+    if s_len > o_len { return Greater;  }
+
```
cuviper: Make these a/b_len too please.

src/bigint.rs
cuviper: I don't know any way to hide it.  This is a public trait implemented for public types -- the result is public.  I suspect if `BigDigit` weren't public, it might stay hidden... but I'm not sure since it's a type alias and not a newtype.

I've tinkered in small ways on the compiler, but nothing like what you're asking for. :)

src/bigint.rs
cuviper: I see, you're counting number of arguments and the result (when present).  OK then.

I see your divide-and-conquer branch, thanks for not stacking that on just yet. :)
We'll get this PR wrapped up first and I'll happily push a new num release for it.

koverstreet: On Tue, Dec 01, 2015 at 04:50:28PM -0800, Josh Stone wrote:
> I see your divide-and-conquer branch, thanks for not stacking that on just yet. :)
> We'll get this PR wrapped up first and I'll happily push a new num release for it.

BTW - I haven't been able to get divide and conquor division to even be as fast
as Knuth, so I doubt I'll be pushing that out any time soon.

It turns out that division of really large numbers is in general (with any
algorithm I've looked at) pretty dependent on multiplication performance - it
looks like well before you get to the point where it's worth doing something
fancier than knuth's algorithm, our current bottleneck is that karatsuba
multiplication doesn't scale that high.

So probably the next thing to do, algorithm wise, is FFT multiplication (or
maybe toom-3, but from what I was reading if you have a good, well optimized
implementation of FFT multiplication it's debatable whether toom-3 is worth it).

But I didn't study any of the math for that, so I'm not sure I'm feeling that
ambitious :)

cuviper:
The new CI failure is in num-macros vs nightly rust -- I'll take a look at that.

I wouldn't even change the size of the public `BigDigit` without calling that a break.  But if we used `TheRealBigDigit = usize` internally (or even ditched giving it a name), that would be fine.

Adding vector conversions for any size u8..u64 sounds intriguing.  The example in `Vec::from_raw_parts` ([1.0](http://doc.rust-lang.org/1.0.0/std/vec/struct.Vec.html#method.from_raw_parts), [nightly](http://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.from_raw_parts)) is a good template.  You'd just have to pad zeros for alignment, tear it apart, restore with new lengths, and fix the endianness as you said.

Ideas for new PRs, not here. ;)

cuviper:
A few review items haven't been addressed yet:

- Sanity checks in the benchmarks during setup, before iteration.
- Add a debug assert for `div_wide` overflow conditions
- Replace public `Add`/`Sub` on slices with private functions.
- Update `cmp_slice` names to a/b_len.

They're all simple things, so I don't want to hold back this PR too long.  If you're already on holiday, maybe I'll go ahead and tweak these myself so we can merge. :)

koverstreet:
On Thu, Dec 10, 2015 at 10:11:57AM -0800, Josh Stone wrote:
> A few review items haven't been addressed yet:
> 
> - Sanity checks in the benchmarks during setup, before iteration.
> - Add a debug assert for `div_wide` overflow conditions
> - Replace public `Add`/`Sub` on slices with private functions.
> - Update `cmp_slice` names to a/b_len.
> 
> They're all simple things, so I don't want to hold back this PR too long.  If you're already on holiday, maybe I'll go ahead and tweak these myself so we can merge. :)

Whoops, sorry, I got distracted with other code. I'll fix those up today :)

koverstreet:
On Thu, Dec 10, 2015 at 10:11:57AM -0800, Josh Stone wrote:
> A few review items haven't been addressed yet:
> 
> - Sanity checks in the benchmarks during setup, before iteration.

Did you want something besides just testing multiply/divide against each other?
Because I added a test that does that.

cuviper:
>
> Did you want something besides just testing multiply/divide against each
> other?
>

That's fine.  I just want a little confidence in what we're benchmarking. :)


> Because I added a test that does that.
>

Right, but that torture test is slow and ignored, and it's not even testing
the same inputs.  That's fine though, since we can easily let the benchmark
sanity-check itself.

koverstreet:
On Thu, Dec 10, 2015 at 01:45:41PM -0800, Josh Stone wrote:
> >
> > Did you want something besides just testing multiply/divide against each
> > other?
> >
> 
> That's fine.  I just want a little confidence in what we're benchmarking. :)
> 
> 
> > Because I added a test that does that.
> >
> 
> Right, but that torture test is slow and ignored, and it's not even testing
> the same inputs.  That's fine though, since we can easily let the benchmark
> sanity-check itself.

The torture test is doing exactly what I thought you meant for what to test in
the benchmark... or did you mean something else?

cuviper:
But it's ignored, for good reason, and not using the same input values as
the benchmark.

Let's not belabor this small point.  If it's still unclear, leave it to me
to add later.

koverstreet:
On Thu, Dec 10, 2015 at 02:07:47PM -0800, Josh Stone wrote:
> But it's ignored, for good reason, and not using the same input values as
> the benchmark.
> 
> Let's not belabor this small point.  If it's still unclear, leave it to me
> to add later.

Ok. Change it however you feel like :)

src/bigint.rs
```diff
@@ -467,146 +530,323 @@ impl Unsigned for BigUint {}
...
+        Greater => {
+            let mut ret = BigUint::from_slice(a);
+            sub2(&mut ret.data[..], b);
+            BigInt::from_biguint(Minus, ret.normalize())
+        },
+        Less    => {
+            let mut ret = BigUint::from_slice(b);
+            sub2(&mut ret.data[..], a);
+            BigInt::from_biguint(Minus, ret.normalize())
+        },
```
cuviper: Surely one of these branches should be `Plus`?  If so, we have insufficient test coverage to notice this, so please add a new test for it too.  That might be best as a direct `sub_sign` unit test, so it's not relying on particular multiplication details to reach this point.

But you had this right before, and multiplication tests still passed, so I wonder how much that calling code is being exercised.  In fact, I just put a `panic!` in `sub_sign`, and it wasn't reached at all until I enabled the torture test.  With that test we can also see the plain failure here.

If we reduce the iterations of that torture test, we can leave it enabled even in debug testing.  Just 1000 iterations still triggers this particular issue.

koverstreet: On Thu, Dec 10, 2015 at 04:06:49PM -0800, Josh Stone wrote:
> >  
> > -        return ll + mm.shl_unit(half_len) + hh.shl_unit(half_len * 2);
> > +    match cmp_slice(a, b) {
> > +        Greater => {
> > +            let mut ret = BigUint::from_slice(a);
> > +            sub2(&mut ret.data[..], b);
> > +            BigInt::from_biguint(Minus, ret.normalize())
> > +        },
> > +        Less    => {
> > +            let mut ret = BigUint::from_slice(b);
> > +            sub2(&mut ret.data[..], a);
> > +            BigInt::from_biguint(Minus, ret.normalize())
> > +        },
> 
> Surely one of these branches should be `Plus`?  If so, we have insufficient test coverage to notice this, so please add a new test for it too.  That might be best as a direct `sub_sign` unit test, so it's not relying on particular multiplication details to reach this point.
>
> But you had this right before, and multiplication tests still passed, so I wonder how much that calling code is being exercised.  In fact, I just put a `panic!` in `sub_sign`, and it wasn't reached at all until I enabled the torture test.  With that test we can also see the plain failure here.

*swear* A bug exactly there is why I overloaded add/subtract for slices...
pretty much reintroduced the same bug when I took it back out. Augh.

Yeah, part of the problem is that karatsuba multiplication doesn't become
worthwhile until the numbers are decently sized.

Does rust support gprof yet? code coverage would be really good to have for code
like this.

> If we reduce the iterations of that torture test, we can leave it enabled even in debug testing.  Just 1000 iterations still triggers this particular issue.

Ok yeah, good idea. Maybe I'll keep a second version that's #ignored that runs
lots more iterations.

cuviper:
@homu r+ fa372e2

cuviper:
Thanks for your hard work!

> Does rust support gprof yet?

Not directly, see rust-lang/rfcs#646.  Apparently kcov works though, like in this [tutorial](https://users.rust-lang.org/t/tutorial-how-to-collect-test-coverages-for-rust-project/650).

PR #135, opened 2015-11-25
Fix float to int cast overflow

### LINKED ISSUE #119: Casting floats to integers does not check overflow
The following code 

``` rust
num::cast::<f64, i32>(1.0e+123f64)
```

will return

``` rust
Some(-2147483648)
```

instead of `None`.

apopiak:
- add impl_to_primitive_float_to_int_or_uint macro with bounds checking
- tests

fix #119

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1136,10 +1136,11 @@ macro_rules! impl_to_primitive_float_to_float {
     )
 }
 
-macro_rules! impl_to_primitive_float_to_int_or_uint {
+macro_rules! impl_to_primitive_float_to_integer {
     ($SrcT:ident, $DstT:ident, $slf:expr) => (
         {
-            if $slf >= $DstT::MIN as $SrcT && $slf <= $DstT::MAX as $SrcT {
+            let t = $slf.trunc();
+            if t >= $DstT::MIN as $SrcT && t <= $DstT::MAX as $SrcT {
                 Some($slf as $DstT)
             } else {
                 None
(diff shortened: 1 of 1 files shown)


apopiak:
feel free to critique

src/traits.rs
```diff
@@ -1136,30 +1136,42 @@ macro_rules! impl_to_primitive_float_to_float {
     )
 }
 
+macro_rules! impl_to_primitive_float_to_int_or_uint {
```
cuviper: (bikeshed) I'd just call it ...`_to_integer`.

src/traits.rs
```diff
@@ -1136,30 +1136,42 @@ macro_rules! impl_to_primitive_float_to_float {
...
+    ($SrcT:ident, $DstT:ident, $slf:expr) => (
+        {
+            if $slf >= $DstT::MIN as $SrcT && $slf <= $DstT::MAX as $SrcT {
```
cuviper: Hmm, how does `as` do rounding/truncation?  I'm concerned that this might reject cases that would have converted to `MIN` or `MAX` just fine, like perhaps `MAX + 0.1`.

(Whether we should do rounding at all is a separate question, but for now we do convert fractional values that are in range of the integer.)

src/traits.rs
```diff
@@ -2448,3 +2460,51 @@ fn integer_decode_f64(f: f64) -> (u64, i16, i8) {
...
+    assert_eq!(Some(normal_f as u16), cast::<f64, u16>(normal_f));
+    assert_eq!(Some(normal_f as u32), cast::<f64, u32>(normal_f));
+    assert_eq!(Some(normal_f as u64), cast::<f64, u64>(normal_f));
+
+    assert_eq!(None, cast::<f64, usize>(small_f));
+    assert_eq!(None, cast::<f64, u8>(small_f));
+    assert_eq!(None, cast::<f64, u16>(small_f));
+    assert_eq!(None, cast::<f64, u32>(small_f));
+    assert_eq!(None, cast::<f64, u64>(small_f));
+}
```
cuviper: Thanks for adding tests!  How about some edge cases near each `MIN` and `MAX`?

src/traits.rs
apopiak: sure, I'll change it.

src/traits.rs
apopiak: `as` truncates when casting from `float` to `int`
If you want to include `MAX + 0.1` we could check for `number > (MIN as float) - 1.0 && number < (MAX as float) + 1.0`
I'm not entirely sure of the edge cases there, though. (e.g what happens if the float inaccuracy comes into play)

src/traits.rs
apopiak: The `MIN` and `MAX` are different for every type! ...

src/traits.rs
cuviper: Ok, since `as` truncates, then try: `let t = f.trunc(); if t >= MIN as float && t <= MAX as float {...}`

src/traits.rs
cuviper: Yes, of course they're different -- that can be macro'ed too.  But such edge case are where it's most important to have testing.  I realize 64-bit integers are tricky though, since MIN/MAX can't even be fully represented in `f64`.

cuviper:
Sorry, I didn't notice you pushed an update for truncation.  That looks good.

On the importance of edge cases, consider something like 2^64, which should be `None` for `u64`, but gives `Some(0)` instead, even with your new code.

```assert_eq!(None, cast::<f64, u64>(64.0.exp2()));```

The true maximum that can be converted is actually less than `std::u64::MAX`, since that value can't even be represented exactly in `f64`.  I think it will be  `0xFFFF_FFFF_FFFF_F800u64`, where in `f64` the 53-bit significand is maxed.  Then `f32` has only a 24-bit significand, so you get `0xFFFF_FF00_0000_0000u64` and `0xFFFF_FF00u32` edge cases.

If only `next_after` were in rust-stable, then we could directly write these limits,  e.g. `64.0.exp2().next_after(0.0)` and `32.0.exp2().next_after(0.0)`.  But running these on nightly confirms the values I gave above.  http://is.gd/xgCyG8

**Comment by apopiak** [comment]
Hey, after his update I did some more testing and realized that my implementation had gaps (which you also found) and I didn't find the time to really dive in again and fix it completely. I'll have another look this week.

**Comment by cuviper** [comment]
If you don't want to hardcode significand-based MIN/MAX values, another possibility is to just check the round-trip.  Something like:

```
let t = f.trunc();
let i = t as $ity;
if i as $fty == t { Some(i) } else { None }
````

But I'd still want those significand edge-cases tested, so we need those values anyway.

cuviper:
We should still fix this, but this PR is long stagnant.  See the new repo and issue here:
https://github.com/rust-num/num-traits/issues/12

PR #106, opened 2015-07-22
Add complex power (.powc()) for T: float.

ticki:

PR #137, opened 2015-12-10
Added powc, powf, log and expf methods for complex numbers

liebharc:
I would like to have a few functions added for complex numbers (powf, log and exp function with arbitrary base). I've provided an implementation with this commit. However it requires more work and discussion and I've added comments to point out the parts which I'm especially unhappy with. Would be nice to get some feedback so that we can improve this pull request first.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -120,7 +120,8 @@ impl<T: Clone + Float> Complex<T> {
     #[inline]
     pub fn exp(&self) -> Complex<T> {
         // formula: e^(a + bi) = e^a (cos(b) + i*sin(b))
-        Complex::new(self.im.cos(), self.im.sin()).scale(self.re.exp())
+        // = from_polar(e^a, b)
+        Complex::from_polar(&self.re.exp(), &self.im)
     }
 
     /// Computes the principal value of natural logarithm of `self`.
@@ -133,7 +134,8 @@ impl<T: Clone + Float> Complex<T> {
     #[inline]
     pub fn ln(&self) -> Complex<T> {
         // formula: ln(z) = ln|z| + i*arg(z)
-        Complex::new(self.norm().ln(), self.arg())
+        let (r, theta) = self.to_polar();
+        Complex::new(r.ln(), theta)
     }
 
     /// Computes the principal value of the square root of `self`.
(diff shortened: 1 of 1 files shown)


src/complex.rs
```diff
@@ -132,6 +133,38 @@ impl<T: Clone + Float> Complex<T> {
         let (r, theta) = self.to_polar();
         Complex::from_polar(&(r.sqrt()), &(theta/two))
     }
+    
+    /// Raises a number to a floating point power.
+    #[inline]
+    pub fn powf(&self, value: T) -> Complex<T> {
+        let (r, theta) = self.to_polar();
+        Complex::from_polar(&(r.powf(value)), &(theta*value))
```
hauleth: There are unneeded parens around `r.powf(value)`.

src/complex.rs
```diff
@@ -132,6 +133,38 @@ impl<T: Clone + Float> Complex<T> {
...
+    }
+    
+    /// Returns the logarithm of the number with respect to an arbitrary base.
+    #[inline]
+    pub fn log(&self, value: T) -> Complex<T> {
+        let (r, theta) = self.to_polar();
+        // Comment for code review:
+        // There is for sure a faster way to get the constant, but I don't know how to do that in this generic context 
+        // Any hints?
+        let e = T::one().exp();
```
liebharc: Any tips how I can get the Euler's constant E for type T?

src/complex.rs
```diff
@@ -132,6 +133,38 @@ impl<T: Clone + Float> Complex<T> {
...
+    }
+    
+    /// Raises a floating point number to a complex power.
+    #[inline]
+    pub fn expb(&self, value: T) -> Complex<T> {
+        // Comment for code review:
+        // Same as value.powf(self) - if it would exist - but I would prefer to have this version
+        // of the function around since I feel like I would look for it in the complex
+        // type. But we would also be able to provide it by creating a trait for real numbers.
+        // Your thoughts?
```
liebharc: Is the function name appropriate? Is it appropriate to have the function here at all?

liebharc:
Thanks for your comment. I saw the merge failure, but there are also 
open questions where I would be happy to get some guidance (what is a 
good way to get Euler's constant of type T and what would be a good name 
for the `expb` function?). Do you have any inputs on those questions so 
that I can incorporate all comments in one push?

Am 4/13/2016 um 1:31 PM schrieb Homu:

hauleth:
@liebharc It is nice that you have started discussion with @homu but this is only bot and I do not think that it will respond.

liebharc:
Oh that's sad :(.

liebharc:
Removing pull request since of missing feedback and I don't want to keep this around as a dead entry any longer.

hauleth:
It is not bad PR. Main reason why there was no feedback is that our work was focused on #164. Reopening

src/complex.rs
cuviper: Since you just want its log, you can change the base.
`e.log(value) = e.ln() / value.ln() = 1.0 / value.ln()`

cuviper: Also, please call the parameter `base`, rather than uninformative `value`.  (ditto everywhere)

src/complex.rs
cuviper: It's a reasonable thing to want.  Not sure about the name, but I can't think of anything better at the moment.  Maybe `expf`, leaving the possibility of `expb` being full `complex^complex` later.

It would help to at least show the formula you're computing.  I get:
`x^(a+bi) = x^a x^bi = x^a e^(b ln(x) i) = from_polar(x^a, b ln(x))`

I think this is mathematically the same as yours, but looks simpler.
(Plain `fn exp` could also be expressed simpler, just `from_polar(e^a, b)`)

cuviper: Silly me, full `complex^complex` should probably be just `pow`.
But I still lean toward `expf` here.

liebharc:
I really appreciate your feedback! Thanks! I'm working on your review comments and will update the pull request as soon as I'm done.

src/complex.rs
liebharc: Very right, the arguments shouldn't be called `value`. Looking through the current API I recognized that the different `pow*` methods use different names for their arguments:
- `PrimInt: pow(self, exp: u32)`
- `Float: powi(self, n: i32), powf(self, n: T)`
- `Rational: pow(&self, expon: i32)`

From this my choice would be `exp`. We could also use this opportunity to clean up the naming and pick one parameter name for all implementations of `pow*`. Would this be a good idea? Parameter names in Rust are from my understanding not part of the method signature and therefore this wouldn't be a breaking chance.

cuviper: `powf(self, exp: T)` sounds fine here.  Let's not bother changing existing functions though -- that's the sort of thing we could clean up if we ever need to make other changes to them.  Otherwise it's unnecessary churn.  You're right though that changing parameter names is not a breaking change.

complex/src/lib.rs
```diff
@@ -150,6 +152,48 @@ impl<T: Clone + Float> Complex<T> {
...
+    pub fn powf(&self, exp: T) -> Complex<T> {
+        let (r, theta) = self.to_polar();
+        Complex::from_polar(&r.powf(exp), &(theta*exp))
+    }
+    
+    /// Returns the logarithm of `self` with respect to an arbitrary base.
+    #[inline]
+    pub fn log(&self, base: T) -> Complex<T> {
+        let (r, theta) = self.to_polar();
+        Complex::new(r.log(base), theta * T::one() / base.ln())
```
cuviper: You don't need the one here, just `theta / base.ln()`.

Please do write a formula comment for consistency on `powf` and `log`, as trivial as they may be.

complex/src/lib.rs
```diff
@@ -150,6 +152,48 @@ impl<T: Clone + Float> Complex<T> {
...
+    }
+    
+    /// Raises `self` to a complex power.
+    #[inline]
+    pub fn pow(&self, exp: Complex<T>) -> Complex<T> {
```
cuviper: Hmm, I do prefer `powc` for this name.  FWIW #106 also started this, and arrived at the same formula as you.  But the implementation and testing was lacking, and the author hasn't responded since, so we can just go with yours.

cuviper:
Please update the PR title and message (first comment) as you go too, because these get immortalized in the merge commit.

complex/src/lib.rs
liebharc: > You don't need the one here, just `theta / base.ln()`.

There are times I feel stupid and sometimes this feeling is well justified. Thanks for the note I will update it soon!

liebharc:
Oh and this is the resource I used to figure out the `powc` formula: http://www.abecedarical.com/zenosamples/zs_complexnumbers.html

I double checked it by comparing some test cases against results obtained with GNU Octave.

cuviper:
Thanks!

@homu r+ 24fcc15

Issue #117, opened 2015-09-17
Implement `From<T>` for `BigInt`/`BigUint` for signed and unsigned integer types (respectively)


hauleth:
As in title. As any integer type can be expressed as `BigInt` and any unsigned integer can be expressed as `BigUint` it would be useful to implement `From<T>` trait for them. It would allow functions to take `T: Into<BigInt>` as parameter and call it with "bare" integers without weird calls and unwraps.

PR #138, opened 2015-12-12
bigint: add From implementations

ollie27:
From all primative unsigned ints to BigUint
From all primative ints to BigInt
From BigUint to BigInt

Closes: #117

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -2412,7 +2412,7 @@ mod biguint_tests {
     use std::i64;
     use std::iter::repeat;
     use std::str::FromStr;
-    use std::u64;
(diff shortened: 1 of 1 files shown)


cuviper:
I'd rather see those benchmarks as simple tests, so they're verified on all targets.  Benchmarks only work on nightlies.  (It's also not very interesting code performance-wise, spending most of its time in the heap allocator for the data Vec.)

Otherwise, the implementation looks great!

ollie27:
I've removed the benchmarks and added some tests.

cuviper:
Well done, thanks!
@homu r+ 9c7c995

PR #139, opened 2015-12-16
bigint: simplify Hash

ollie27:
There cannot be any leading zeros in a BigUint so just derive Hash which will just hash the Vec directly.
Add Hash to Sign so we can derive it for BigInt as well.


cuviper:
Nice -- it's quite a bit faster too.  I wrote a quick benchmark (which I'll push after) that goes from 525us to 138us on the derived Hash.

@homu r+ e152bbc

hauleth:
:+1:

PR #142, opened 2015-12-18
Add bit manipulation algorithms to PrimInt

### LINKED ISSUE #80: All bit twiddling functions in num::traits::PrimInt should take or return usize instead of u32
The trait `num::traits::PrimInt` mandates that any type implementing it must also implement `Shl<usize, Output=Self>` and `Shr<usize, Output=Self>` (in other words, `PrimInt` mandates that the shift amount be of type `usize`).

Thus all bit twiddling functions (`trailing_zeros`, `rotate_left`, `count_zeros` etc.) in `PrimInt` should take or return a `usize` instead of `u32` as the number of bits for the sake of consistency with what the implementation restrictions on `Shl` and `Shr` mandate.

### LINKED ISSUE #141: Raw bit-preserving conversion from signed/unsigned to unsigned/signed
I'm implementing some bit manipulation algorithms on top of `PrimInt` and there are some situations in which I need to get an unsigned representation of the Integer type to operate on the bits directly using wrap around unsigned arithmetic. 

Is there a way to:
- obtain the unsigned integer type that has the same size as the current integer type
- cast the current type to its associated unsigned integer type preserving the bits (i.e. transmute)? 

The reciprocal operations (going from unsigned to signed) are also required.

gnzlbg:
This adds to the `PrimInt` trait:
- two associated types: `Signed`/`Unsigned`, which contains a type of the same `mem::size_of` as `Self`, but with the corresponding signedness, and two functions `bits_to_signed/unsigned -> Signed/Unsigned`; fixes #141 .
- modifies the requirements of `Shl``Shr`to require`u32`instead of`usize`, fixes #80 .
- some bit manipulation algorithms, mainly adapted from the [C++ bitops proposal](https://github.com/fmatthew5876/stdcxx-bitopsstd). 

Most of the algorithms have sub-optimal implementations since:
- the fastest implementation is to use either a compiler intrinsic or a platform-dependent intrinsic (e.g. extracting/depositing bits with Intel's BMI2.0 is much faster than any hand-rolled implementation).
- I wanted to get correct implementations working first, add lots of tests, and optimize later,
- some optimizations are size dependent, and while some times mem::size_of is used to help the compiler, they are better provided in different implementations for each type.

TODOs:
- revert `Shl<u32>` to `Shl<usize>`
- gather consensus around whether we need the associated types Signed/Unsigned or not and about how these should be implemented (maybe in a different trait that requires `PrimInt` to avoid breaking its API). 
- TODO: find real world examples of the associated types or remove them
- TODO: document the intrinsics that should be used for each algorithm
- TODO: add assertions
- TODO: where to put the bit manipulation algorithms
- TODO: many more tests

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1584,7 +1584,7 @@ pub trait PrimInt
     /// ```
     fn align_down(self, alignment: u32) -> Self;
 
-    /// Outer Perfect Shuffle
+    /// Outer Perfect Shuffle of `self`
     ///
     /// See also [Hacker's Delight: shuffling bits](http://icodeguru.com/Embedded/Hacker's-Delight/047.htm).
     ///
@@ -1602,7 +1602,7 @@ pub trait PrimInt
     /// ```
     fn outer_perfect_shuffle(self) -> Self;
 
-    /// Outer Perfect Unshuffle
+    /// Outer Perfect Unshuffle of `self`
     ///
     /// See also [Hacker's Delight: shuffling bits](http://icodeguru.com/Embedded/Hacker's-Delight/047.htm).
     ///
@@ -1620,7 +1620,7 @@ pub trait PrimInt
     /// ```
     fn outer_perfect_unshuffle(self) -> Self;
 
-    /// Inner Perfect Shuffle
+    /// Inner Perfect Shuffle of `self`
     ///
     /// See also [Hacker's Delight: shuffling bits](http://icodeguru.com/Embedded/Hacker's-Delight/047.htm).
     ///
@@ -1640,7 +1640,7 @@ pub trait PrimInt
         self.reverse_bit_groups((mem::size_of::<Self>() * 8 / 2) as u32, 1).outer_perfect_shuffle()
     }
 
-    /// Inner Perfect Unshuffle
+    /// Inner Perfect Unshuffle of `self`
     ///
     /// See also [Hacker's Delight: shuffling bits](http://icodeguru.com/Embedded/Hacker's-Delight/047.htm).
     ///
@@ -1659,6 +1659,46 @@ pub trait PrimInt
     fn inner_perfect_unshuffle(self) -> Self {
         self.outer_perfect_unshuffle().reverse_bit_groups((mem::size_of::<Self>() * 8 / 2) as u32, 1)
     }
+
+    /// Parallel bits deposit of `mask` into `self`
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num::traits::PrimInt;
+    ///
+    /// let n  = 0b1011_1110_1001_0011u16;
+    ///
+    /// let m0 = 0b0110_0011_1000_0101u16;
+    /// let s0 = 0b0000_0010_0000_0101u16;
+    ///
+    /// let m1 = 0b1110_1011_1110_1111u16;
+    /// let s1 = 0b1110_1001_0010_0011u16;
+    ///
+    /// assert_eq!(n.parallel_bits_deposit(m0), s0);
+    /// assert_eq!(n.parallel_bits_deposit(m1), s1);
+    /// ```
+    fn parallel_bits_deposit(self, mask_: Self) -> Self; 
+
+    /// Parallel bits extract of `mask` from `self`
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use num::traits::PrimInt;
+    ///
+    /// let n  = 0b1011_1110_1001_0011u16;
+    ///
+    /// let m0 = 0b0110_0011_1000_0101u16;
+    /// let s0 = 0b0000_0000_0011_0101u16;
+    ///
+    /// let m1 = 0b1110_1011_1110_1111u16;
+    /// let s1 = 0b0001_0111_0100_0011u16;
+    ///
+    /// assert_eq!(n.parallel_bits_extract(m0), s0);
+    /// assert_eq!(n.parallel_bits_extract(m1), s1);
+    /// ```
+    fn parallel_bits_extract(self, mask_: Self) -> Self;
 }
 
 macro_rules! prim_int_impl {
@@ -1817,6 +1857,41 @@ macro_rules! prim_int_impl {
                     x = x ^ t ^ (t << 16);
                 }
                 return x;
+            }
+
+            fn parallel_bits_deposit(self, mask_: Self) -> Self
+            {
+                let mut res = <Self as Zero>::zero();
+                let mut mask = mask_;
+                let mut bb = <Self as One>::one();
+                loop {
+                    if mask == <Self as Zero>::zero() {
+                        break;
+                    }
+                    if (self & bb) != <Self as Zero>::zero() {
+                        res = res | (mask & mask.wrapping_neg());
+                    }
+                    mask = mask & (mask - <Self as One>::one());
+                    bb = bb + bb;
+                }
+                res
+            }
+
+            fn parallel_bits_extract(self, mask_: Self) -> Self {
+                let mut res = <Self as Zero>::zero();
+                let mut mask = mask_;
+                let mut bb = <Self as One>::one();
+                loop {
+                    if mask == <Self as Zero>::zero() {
+                        break;
+                    }
+                    if self & mask & (mask.wrapping_neg()) != <Self as Zero>::zero() {
+                        res = res | bb;
+                    }
+                    mask = mask & (mask - <Self as One>::one());
+                    bb = bb + bb;
+                }
+                res
             }
         }
     )
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1102,7 +1102,45 @@ pub trait PrimInt
        self ^ (self + <Self as One>::one())
    }
 
-   fn general_reverse_bits(self, subword_bits: u32, group_subwords: u32) -> Self;
+   /// Reverses the bits of `self` by `subword_bits` and `group_subwords`:
+   ///
+   /// - `subword_bits`: the bits will be reversed in grous:
+   ///   1 (single bits), 2 (pair-wise), 4 (nibbles)
+   /// - `group_subwords`: the subword size is 8 bits: `mem::size_of::<u8>()`,
+   ///   the bits will be reversed within each subword.
+   ///
+   /// # Examples
+   ///
+   /// ```
+   /// use num::traits::PrimInt;
+   ///
+   /// let n = 0b0101_1101_1010_0101_u16;
+   ///
+   /// // Single bits:
+   /// let s0 = 0b1010_0101_1011_1010u16;
+   /// assert_eq!(n.reverse_bit_groups(1, 1), s0);
+   ///
+   /// // Bit pairs:
+   /// let s1 = 0b0101_1010_0111_0101u16;
+   /// assert_eq!(n.reverse_bit_groups(2, 1), s1);
+   ///
+   /// // Bit nibbles:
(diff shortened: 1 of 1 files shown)


gnzlbg:
Since this is my first contribution to a Rust project, I am submitting this PR in the current unfinished form to gain some feedback:
- Is `PrimInt` the right place to add bitwise manipulation algorithms for primitive integer types?
- Is the way I'm using cast(1).unwrap() in the `PrimInt` trait ok? Or should I implement those methods in the impls to avoid the cast and unwrap? (In the impls I can just: "1 as Self", but I cannot do that in the Trait)
- Is there a "Transmutable" trait or similar? I require the Signed and Unsigned associated types to be PrimInts as well in the PrimInt trait, but I would like to strengthen this requirement to say that these types must have the same mem::size_of as Self. Since transmute has this requirement I would think that transmute would be constrained by a magic Transmutable trait that checks for this but I cannot find it.

src/traits.rs
```diff
@@ -671,7 +701,7 @@ pub trait PrimInt
     /// ```
     /// use num::traits::PrimInt;
     ///
-    /// let n = 0b01001100u8;
+    /// let n = 0b0100_1100u8;
```
hauleth: There is no need IMHO to introduce separator in quartets of bits. Octets are readable enough for me.

gnzlbg:
@hauleth those were unintentional sorry about that, i've reverted them. In the new functions I've used blocks of 4 bit since octets are too long for my eyes (with 4 bit blocks I see that there are 4 elements, but 8 bits are long enough that I have to count them to make sure that there aren't 7 or 9). 

Maybe I should still change the new functions to 8bit blocks for the documentation to follow a consistent style? Am I the only one with that problem when reading 8bit blocks?

cuviper:
>two associated types: Signed/Unsigned, which contains a type of the same mem::size_of as Self, but with the corresponding signedness, and two functions bits_to_signed/unsigned -> Signed/Unsigned; fixes #141 .

I'm still looking for the motivation for this.  You did use `Unsigned` in the impl of `general_reverse_bits`, but it was unnecessary there since that's not generic code -- you have `$AU`.

>modifies the requirements of Shl`Shrto requireu32instead ofusize`, fixes #80 .

Let's not make any breaking changes at this time.

>Is PrimInt the right place to add bitwise manipulation algorithms for primitive integer types?

Maybe.  It sort of fits the existing methods like `count_ones`, at least.

>Is the way I'm using cast(1).unwrap() in the PrimInt trait ok? Or should I implement those methods in the impls to avoid the cast and unwrap? (In the impls I can just: "1 as Self", but I cannot do that in the Trait)

This is what `Zero` and `One` are for.  In the impls you can probably just use `1` alone and let it infer the type, but `1 as Self` is ok when it can't be inferred.

>Is there a "Transmutable" trait or similar? I require the Signed and Unsigned associated types to be PrimInts as well in the PrimInt trait, but I would like to strengthen this requirement to say that these types must have the same mem::size_of as Self. Since transmute has this requirement I would think that transmute would be constrained by a magic Transmutable trait that checks for this but I cannot find it.

This came up in the internals forum recently [asking for a `SameSize` trait](https://internals.rust-lang.org/t/trait-for-enforcing-type-size-constraints/2901), but such a thing doesn't exist yet.  The checks done by `transmute` are deep compiler magic not available to the likes of us. :) See [`check_intrinsics`](https://github.com/rust-lang/rust/blob/master/src/librustc_trans/trans/intrinsic.rs#L101).

Maybe there's a way to constrain `Self::Unsigned::Signed = Self::Signed` and vice versa - I'm not sure.

src/traits.rs
```diff
@@ -848,11 +878,491 @@ pub trait PrimInt
...
+   /// let n = 0b0101_1111u8;
+   /// let s = 0b0011_1111u8;  
+   ///
+   /// assert_eq!(n.mask_trailing_ones_and_least_significant_zero(), s);
+   /// ```
+   fn mask_trailing_ones_and_least_significant_zero(self) -> Self {
+       self ^ (self + cast(1).unwrap())
+   }
+
+   fn general_reverse_bits(self, subword_bits: u32, group_subwords: u32) -> Self;
```
cuviper: Since this is undocumented, I guess it's meant to be an implementation detail, but all functions in a pub trait are exposed.  You'll need to pull this out and implement it separately.

src/traits.rs
```diff
@@ -848,11 +878,491 @@ pub trait PrimInt
...
+   ///
+   /// let n1 = 0b1011_0010_1010_1001u16;
+   /// let s1 = 0b1001_1010_0010_1011u16;
+   /// assert_eq!(n1.reverse_bit_nibbles(), s1);
+   /// ```
+   fn reverse_bit_nibbles(self) -> Self {
+       self.general_reverse_bits(4, 1)
+   }
+
+   fn reverse_byte_groups(self, bytes_per_block: u32, blocks_per_group: u32) -> Self {
```
cuviper: Ditto on public implementation details.

src/traits.rs
```diff
@@ -848,11 +878,491 @@ pub trait PrimInt
...
+   ///
+   /// let n = 0b1011_0010u8;
+   /// let s = 0b1011_0010u8;  
+   /// assert_eq!(n.reverse_bytes(), s);
+   ///
+   /// let n1 = 0b1011_0010_1010_1001u16;
+   /// let s1 = 0b1010_1001_1011_0010u16;
+   /// assert_eq!(n1.reverse_bytes(), s1);
+   /// ```
+   fn reverse_bytes(self) -> Self {
```
cuviper: Isn't this just like `swap_bytes`?

src/traits.rs
```diff
@@ -848,11 +878,491 @@ pub trait PrimInt
...
+   /// ```
+   /// use num::traits::PrimInt;
+   ///
+   /// let n = 0b1011_0010u8;
+   /// assert_eq!(n.test_bit(7), true);
+   /// assert_eq!(n.test_bit(6), false);
+   /// assert_eq!(n.test_bit(5), true);
+   /// ```
+   fn test_bit(self, bit: u32) -> bool {
+        self & (cast::<_, Self>(1).unwrap() << bit) > cast::<_, Self>(0).unwrap()
```
cuviper: I would test not equal to zero.

src/traits.rs
```diff
@@ -900,11 +1410,62 @@ macro_rules! prim_int_impl {
...
+                      if width > i && (k & s > 0) {
+                          x = ((x & (l as Self::Unsigned)) << (s as Self::Unsigned))
+                              | ((x & (r as Self::Unsigned)) >> (s as Self::Unsigned));
+                      }
+                      println!("up1({}): x = {:b}", i, x);
+                  };
+
+                  up1(1, 8, 0x00FF00FF00FF00FFusize, 0xFF00FF00FF00FF00usize);
+                  up1(2, 16, 0x0000FFFF0000FFFFusize, 0xFFFF0000FFFF0000usize);
+                  up1(4, 32, 0x00000000FFFFFFFFusize, 0xFFFFFFFF00000000usize);
```
cuviper: All of these constants need to be `u64`, since `usize` is smaller on 32-bit platforms.
(And of course the prints need to be removed before we can merge.)

src/traits.rs
```diff
@@ -656,14 +656,44 @@ pub trait PrimInt
...
+    /// # Examples
+    ///
+    /// ~~~
+    /// use num::traits::PrimInt;
+    ///
+    /// assert_eq!((8i32).bits_to_unsigned().bits_to_signed(), 8i32);
+    /// assert_eq!((-1i32).bits_to_unsigned().bits_to_signed(), -1i32);
+    /// ~~~
+    fn bits_to_signed(self) -> Self::Signed;
+
```
cuviper: I'm still not sure we want these, but if so, I'd call them `to_unsigned` and `to_signed`.

gnzlbg:
@cuviper modifies the requirements of ShlShrto requireu32instead ofusize, fixes #80 .
>> 
>
> Let's not make any breaking changes at this time.

This means that in the algorithms taking e.g. `bit: u32` (like `set_bit`) I either need to take `bit: usize` or cast from `u32` to `usize` in the implementation (either using `cast` within the trait or by moving the implementations to the macro and using `as`). 

Any recommendations on how to proceed with these?

gnzlbg:
> This is what Zero and One are for. In the impls you can probably just use 1 alone and let it infer the type, but 1 as Self is ok when it can't be inferred.

@cuviper  I've switched to `<Self as One>::one()` in the trait. For some reason Rust complain if I just use `Self::one()`.

src/traits.rs
gnzlbg: Fixed/

src/traits.rs
gnzlbg: Fixed.

src/traits.rs
gnzlbg: I've renamed them for now, will remove them later if the consensus says that we do not need them in `PrimInt` (I guess I will then propose another trait that implement these and requires `PrimInt`, which could be the way we end up doing this anyways even if consensus agrees we need them to avoid breaking the API). I've added a TODO to the original issue to track this.

src/traits.rs
gnzlbg: I've documented this, i think it is useful.

src/traits.rs
gnzlbg: I've renamed the function to `reverse_bit_groups` and documented it, it is very useful. I am not sure about the name though.

src/traits.rs
gnzlbg: Yes it is, I've added a note, an assert, and forwarded it to `swap_bytes` since it should use the llvm `swapbs` intrinsics which is faster. 

I don't know if I should remove this function or leave it though. For consistency with the other bit reversal algorithms it makes sense for it to exist and users might wonder why not, but those looking for the CPU / LLVM intrinsic will look for `swap_bytes`. However the other reversal algorithms are not called `swap_bits` or similar. 

Need to give more thought to this.

cuviper:
I know the `u32`/`usize` disparity is unfortunate.  It's on my list of things that we should clean up, whenever we decide to bump semver and make breaking changes -- I just want that to be a rare event.  Anyway, I'm not sure why you need `cast` in this case; since the two types are known, `as` is fine.

I'm hesitating that this is all a lot to add.  Right now `PrimInt` is merely mapping the common inherent methods of the primitives.  Since you're modelling these functions after the C++ bitops proposal, maybe this would make more sense in your own `bitops` crate?  What would we need to enable you to write that generically?  I offered `arithmetic_shr` and `logical_shr` in #141 -- anything else?

gnzlbg:
> Anyway, I'm not sure why you need cast in this case; since the two types are known, as is fine.

A lot of algorithms are implemented in the trait. As mentioned above I can either use cast there, or move their implementation into the impl macro and use `as` there. Since I don't know much rust I was having doubts about which is the better way of doing this. My gut feeling tells me that `as` is probably faster/better than using `cast` (which returns an option that cannot be optimized for integers) so I guess the right solution is to just move these into the macro impl. Is that so or is there a better way?

gnzlbg:
So i've just added the alignment, shuffling, and deposit/extract algorithms. The only thing I needed besides arithmetic shifts is wrapping negation and have opened an issue for that. With those two things I'm pretty certain I could put the bit manipulation algorithms into their own trait that requires only PrimInt.

cuviper:
AIUI, you've published this in https://github.com/gnzlbg/bitwise, so I'm going to close here.

PR #143, opened 2015-12-19
Add linspace(), a sequence of evenly spaced floats

bluss:
Linspace produces `n` elements between a and b, both ends inclusive.

We compute `step = (b - a)/(n - 1)` and each element as `a + i * step`
where `i` is the iteration counter.

This method is equivalent to what Python's numpy.linspace does, and it
also makes sure the iterator can be double ended (an accumulation would
not be reversible due to rounding errors).

This iterator has previously appeared in itertools, but it has a better
home in num.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -10,7 +10,9 @@
 
 //! External iterators for generic mathematics
 
-use {Integer, Zero, One, CheckedAdd, ToPrimitive};
+use {Float, Integer, Zero, One, CheckedAdd, ToPrimitive};
+use traits::cast;
+
 use std::ops::{Add, Sub};
 
 /// An iterator over the range [start, stop)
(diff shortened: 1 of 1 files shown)


bluss:
I suppose if this doesn't fit in num, there could be room for a new crate for the overlap of common iterator and related basics and numerics.

bluss:
Build failure is because of num_macros, fixed in separate PR.

bluss:
numpy has a bells and whistles api for linspace, using default arguments. It's actually not impossible to extend this with a semi-builder method `.exclude_end()` that simply recomputes the step. The unused parts should optimize away. The only awkward part is that calling .exclude_end() on an iterator that's been modified is not wrong, but likely not something you want.

cuviper:
Hmm, do you think this could be written more generically than just `Float` types?  For instance, numpy allows a different `dtype` output, although the step is always a float:

```
>>> numpy.linspace(1, 10, 12, dtype=int, retstep=True)
(array([ 1,  1,  2,  3,  4,  5,  5,  6,  7,  8,  9, 10]), 0.8181818181818182)
```

In `num`, I could imagine wanting more perfect precision with `Ratio`, for instance.  Perhaps it could be something like `Linspace<Value, Step=Value>` and let the caller choose?  With some appropriate arithmetic and casting traits required, of course.

Maybe it's more Rustic to let `Linspace` deal only in the step type, and the user can map it down to truncate if desired.  But `Ratio` still seems like a reasonable choice, instead of only `Float`.

Is it guaranteed that the endpoint will be reached precisely?  I would think not, since the step is subject to rounding error, as you noted about reverse iteration.

The builder/iterator conflict you mention can be solved by separating into two types, `Linspace` as the builder and `LinspaceIterator` to iterate, with `IntoIterator` between.  And perhaps a plain `fn iter()` as shorthand for when `into_iter()` isn't implicit.

bluss:
Supporting ratio sounds good. I also pondered the builder + into iterator possibility, but both that and not producing float directly makes for a very inconvenient interface.

bluss:
I think it's an interesting project, but I won't work on it on num for now.

PR #144, opened 2015-12-19
Fix num_macros for syntax/syntax_ext crate split

bluss:


bluss:
I don't even know what syntax extensions num has, but this fixes the build.. :smile:

cuviper:
It allows `#[derive(FromPrimitive)]`.  I have no idea if anyone actually uses this, but as long as the maintenance burden isn't too high, fine. :)

I just pushed a fix needed for the doctest too, so let's see if homu will merge the way I think...

@homu r+ 4d23700

PR #145, opened 2015-12-23
bigint: fix float conversions

ollie27:
The default implementations of to_f32, to_f64, from_32 and from_f64 are
limited to numbers fitting in i64 but BigInt can of course be much bigger.
The default to_f32 also has double rounding.
from_f32 and from_f64 have undefined behaviour if the float is too big to
fit in an i64.

This fixes these issues and keeps the rounding consistant with other float
to int conversions.

Also add ToBigUint and ToBigInt implementations for f32 and f64.

Currently this returns None if the BigInt is too big for to_f32 and to_f64 but it might make more sense to return Some(INFINITY) but I've decided to match f64::to_f32() for now.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -1178,16 +1178,16 @@ impl ToPrimitive for BigUint {
             0 => Some(f32::zero()),
             1 => Some(self.data[0] as f32),
             len => {
-                // prevent overflow of exponant
+                // this will prevent any overflow of exponent
                 if len > (f32::MAX_EXP as usize) / big_digit::BITS {
                     None
                 } else {
-                    let exponant = (len - 2) * big_digit::BITS;
+                    let exponent = (len - 2) * big_digit::BITS;
                     // we need 25 significant digits, 24 to be stored and 1 for rounding
                     // this gives at least 33 significant digits
                     let mantissa = big_digit::to_doublebigdigit(self.data[len - 1], self.data[len - 2]);
                     // this cast handles rounding
-                    let ret = (mantissa as f32) * 2.0.powi(exponant as i32);
+                    let ret = (mantissa as f32) * 2.0.powi(exponent as i32);
                     if ret.is_infinite() {
                         None
                     } else {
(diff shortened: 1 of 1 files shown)


hauleth:
For me seems ok. @cuviper what you think?

src/bigint.rs
```diff
@@ -1168,6 +1170,71 @@ impl ToPrimitive for BigUint {
...
     }
+
+    // `DoubleBigDigit` size dependent
+    #[inline]
+    fn to_f32(&self) -> Option<f32> {
+        match self.data.len() {
+            0 => Some(f32::zero()),
+            1 => Some(self.data[0] as f32),
+            len => {
+                // prevent overflow of exponant
```
cuviper: *exponent

src/bigint.rs
```diff
@@ -1168,6 +1170,71 @@ impl ToPrimitive for BigUint {
...
+                    let needed = (f64::MANTISSA_DIGITS as usize) + 1;
+                    let bits = (2 * big_digit::BITS) - (mantissa.leading_zeros() as usize);
+                    if needed > bits {
+                        let diff = needed - bits;
+                        mantissa <<= diff;
+                        exponant -= diff;
+                        let mut x = self.data[len - 3];
+                        x >>= big_digit::BITS - diff;
+                        mantissa |= x as u64;
+                    }
```
cuviper: How about just padding to a full 64-bit value?  Something like:

```
let shift = mantissa.leading_zeros();
if shift > 0 {
    mantissa <<= shift;
    mantissa |= self.data[len - 3] as u64 >> (big_digit::BITS - shift);
    exponent -= shift;
}
```

src/bigint.rs
```diff
@@ -1184,6 +1251,41 @@ impl FromPrimitive for BigUint {
     fn from_u64(n: u64) -> Option<BigUint> {
         Some(BigUint::from(n))
     }
+
+    #[inline]
+    fn from_f32(n: f32) -> Option<BigUint> {
+        BigUint::from_f64(n as f64)
+    }
```
cuviper: `FromPrimitive` already does this, no?

cuviper:
A few small line notes, otherwise it looks good to me.  Thanks especially for thorough tests!

ollie27:
I've addressed the comments, let me know if you want me to squash the commits.

cuviper:
Looks good, thanks.

@homu r+ 2c235d2

PR #149, opened 2015-12-29
Add complex `i` constant function

hauleth:


src/complex.rs
```diff
@@ -59,6 +59,14 @@ impl<T: Clone + Num> Complex<T> {
     }
 }
 
+impl<T: Clone + Num> Complex<T> {
+    /// Returns `i`
+    #[inline]
+    pub fn i() -> Complex<T> {
+        Self::new(T::zero(), T::one())
+    }
+}
```
cuviper: Why did you make this a separate `impl` block from the one above?
I'd also be a little more descriptive, like "Returns the imaginary unit."

hauleth: Accidentally as I firstly wrote that for types that are `Clone + Zero + One`, not `Clone + Num`.

cuviper: @homu r+

cuviper:
@homu r+

PR #152, opened 2016-01-06
Improve pow speed for expensive types by multiplying references not cloned values

dten:
pow can be made cheaper for types that are expensive to clone such as BigUInt.

includes a benchmark that goes from
`test pow_bench         ... bench: 159,222,506 ns/iter (+/- 8,973,716)`
to
`test pow_bench         ... bench: 140,754,320 ns/iter (+/- 7,129,689)`

not sure if this is a breaking change though if Mul for references isn't implemented?

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -192,7 +192,7 @@ fn hash(b: &mut Bencher) {
 #[bench]
 fn pow_bench(b: &mut Bencher) {
     b.iter(|| {
-        let upper = 250_usize;
+        let upper = 100_usize;
         for i in 2..upper + 1 {
             for j in 2..upper + 1 {
                 let i_big = BigUint::from_usize(i).unwrap();
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -188,3 +188,16 @@ fn hash(b: &mut Bencher) {
         assert_eq!(h.len(), v.len());
     });
 }
+
+#[bench]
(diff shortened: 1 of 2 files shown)


cuviper:
Yes, I would consider it a breaking change, but probably a worthwhile to plan on doing.  I'll add it to the tracker.

All the primitives do implement ops by reference, even though they can just copy.  I expect most complicated types which care about cloning would also implement ops by reference.  So it's a pretty reasonable expectation, but still a breaking change.

While I'm looking at this code, I see a bit of optimization I can do around powers of two in the exponent.  I tried it briefly, and it does show a modest improvement in your benchmark.  This is orthogonal to cloning changes though.  If you don't mind, I'd like to commit just your benchmark for now, with my tweak on top, then we'll leave the rest open for when we make breaking changes.

dten:
Sure thing. Do you need me to do anything?

cuviper:
No, I can pick it out with authorship preserved.  You'll probably have to fix up merge conflicts afterward though.  I'll try to get to this tonight, but no rush on your part since we'll be waiting for a breaking-change window anyway.

dten:
Sorry I managed to mess this up. My git skills are weak :-1:

cuviper:
Can you reopen now that you restored your branch?  If not, a new PR is fine.  This may sit for a while though, so you may want to use a dedicated name in your tree besides "master".

PR #153, opened 2016-01-07
Improve pow speed for expensive types by multiplying references not cloned values (reopen)

dten:
Resubmission of https://github.com/rust-num/num/pull/152
Couldn't reopen because I broke the branch

Speed up on pow_bench

```
   test pow_bench         ... bench:   9,579,445 ns/iter (+/- 560,439)
-> test pow_bench         ... bench:   6,820,137 ns/iter (+/- 760,415)
```

### DIFF
--- a/traits/src/pow.rs
+++ b/traits/src/pow.rs
@@ -12,11 +12,14 @@ use {One, CheckedMul};
 /// assert_eq!(pow(6u8, 3), 216);
 /// ```
 #[inline]
-pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T {
+pub fn pow<T: Clone + One + Mul<T, Output = T>>(mut base: T, mut exp: usize) -> T
+where
+    for<'a> &'a T: Mul<&'a T, Output = T>,
+{
     if exp == 0 { return T::one() }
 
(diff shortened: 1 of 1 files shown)


cuviper:
Great.  As noted in #152, we'll hold this until we're ready to make breaking changes.

vks:
I think this could be even more efficient by using `x = x * &y` instead of `x = &x * &y`.

dten:
@vks i suspect you're right :+1:

cuviper:
Do you want to move this PR to the `next` branch?  That's where I'm (slowly) making breaking changes, although I'm not sure when that will get around to a release.

dten:
That sounds the best place for it yes please

cuviper:
Great, let's go!

bors r+

cuviper:
Hmm, in `next` we have a default implementation of `Float::powi` using this `pow`, but `Float` doesn't have reference operators.  But I suppose we could expand those requirements...

cuviper:
Oh, but indirect trait requirements don't propagate well.  (rust-lang/rust#20671)

I guess we can just duplicate the `powi` implementation by value for this case.  It's not *that* much code.

cuviper:
Merged in #334, thanks!

PR #154, opened 2016-01-08
Make `rand` and `rustc-serialize` dependencies optional.

briansmith:
Previously, the `rand` and `rustc-serialize` dependencies were optional
except they were required for the `bigint` feature.

Make the dependency on the `rand` crate optional in all cases.
including when the `bigint` feature is selected. Some of the tests for
the bigint feature are randomized so, while `rand` is now an optional
dependency, it is a non-optional dev-dependency.

Similarly, make the dependency on the `rustc-serialize` crate optional
in all cases, including when the `bigint` feature is selected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,12 +17,17 @@ complex, rational, range iterators, generic integers, and more!
 rustc-serialize = { version = "0.3.13", optional = true }
 rand = { version = "0.3.8", optional = true }
 
+[dev-dependencies]
+# Some tests of non-rand functionality still use rand because the tests
+# themselves are randomized.
+rand = { version = "0.3.8" }
+
 [features]
 
 complex = []
 rational = []
-bigint = ["rustc-serialize", "rand"]
-default = ["complex", "rational", "bigint"]
+bigint = []
+default = ["bigint", "complex", "rand", "rational", "rustc-serialize"]
 
 [[bench]]
 name = "bigint"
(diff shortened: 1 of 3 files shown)


briansmith:
The Nightly build fails because of rust-lang/rust#30713, which should be fixed in the next Rust Nightly. With this commit, packages depending on `num` can avoid that breakage by using `default-features = false` and specifying `features = [<anything but "rand">]` in the dependencies.

cuviper:
I think this is good.  My only concern is that someone using just feature bigint will now lose functionality, so that's a breaking change.  But since this isn't the default case, I'm leaning toward accepting that.  What do you think @hauleth?

hauleth:
I don't see how one could loose functionality by using only `bignum`. Rather I can see how one could loose functionality when providing `no-default` and forgetting to add `rand` to the feature list. So yes. I would consider it somewhat breaking change, but I do not think that impact will be wide.

cuviper:
OK, you described what I meant.  We'll go ahead then.

@homu r+ 9e3e855

cuviper:
Grr, still no new published nightly to fix rand.
Locally with `rustc 1.7.0-nightly (8f11a9ef4 2016-01-03)` looks fine.

hauleth:
Yeah, I am also waiting for that fix. Today should be new nightly build. It is strange. 

--
Łukasz Jan Niemier

Dnia 8 sty 2016 o godz. 18:52 Josh Stone <notifications@github.com> napisał(a):

PR #157, opened 2016-01-15
remove stale #[deriving(Zero)] docs

### LINKED ISSUE #156: Zero's documentation is out of date
It says Zero can be automatically derived using `#[deriving]` which is out of date syntax. And probably misleading anyway, since you'll need a compiler plugin to use such a derive.

durka:
Fixes #156.


hauleth:
@homu r+

PR #158, opened 2016-01-17
Add LowerHex and UpperHex formatting

anchovieshat:
These were mysteriously missing from BigUint.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -242,6 +242,18 @@ impl fmt::Display for BigUint {
     }
 }
 
+impl fmt::LowerHex for BigUint {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.to_str_radix(16))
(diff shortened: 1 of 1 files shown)


anchovieshat:
Collections were not stabilized in ```1.0.0```, causing the build failure. I could remove ```UpperHex```, or write a little extra code to make the characters uppercase.

cuviper:
Looks good -- how about a quick test?

anchovieshat:
What is the best way to test this? Verify that the resulting string is the same as what ```from_str_radix``` would have returned? I could check that ```UpperHex``` and ```LowerHex``` return the same value?

cuviper:
I don't think it needs very rigorous or exhaustive testing -- just
demonstrate things that wouldn't have worked before.  Maybe try a few
`format!` calls compared with expected hex output.

cuviper:
Great, thanks!

@homu r+ a89a2e8

PR #160, opened 2016-01-20
bigint: Improve fmt traits

ollie27:
* Make them take into account formatting parameters.
* ~~Make Debug match Display.~~

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -239,31 +239,31 @@ impl Default for BigUint {
 
 impl fmt::Display for BigUint {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_str_radix(10))
+        f.pad_integral(true, "", &self.to_str_radix(10))
     }
 }
 
 impl fmt::LowerHex for BigUint {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_str_radix(16))
+        f.pad_integral(true, "0x", &self.to_str_radix(16))
     }
 }
 
 impl fmt::UpperHex for BigUint {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_str_radix(16).to_ascii_uppercase())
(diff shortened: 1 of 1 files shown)


cuviper:
The general formatting changes are good, but I think `Debug` should be left alone.  From the [docs](http://doc.rust-lang.org/std/fmt/#fmtdisplay-vs-fmtdebug):

> fmt::Debug implementations should be implemented for all public types. Output will typically represent the internal state as faithfully as possible. The purpose of the Debug trait is to facilitate debugging Rust code. In most cases, using #[derive(Debug)] is sufficient and recommended.

ollie27:
As far as facilitating debugging Rust code goes I think it will be more useful for people using `BigInt` to see the actual integer it represents rather than how it is represented which contains a lot of noise. For example the output from `assert_eq!(BigInt::zero(), BigInt::one())`:

``assertion failed: `(left == right)` (left: `BigInt { sign: NoSign, data: BigUint { data: [] } }`, right: `BigInt { sign: Plus, data: BigUint { data: [1] } }`)``
vs.
``assertion failed: `(left == right)` (left: `0`, right: `1`)``

It will only get worse if you start using `BigInt`s inside other data structures.

This matches the `Debug` output of the primitive ints which is what I was trying to stay consistent with.

cuviper:
Asserts don't need to be pretty.  Primitive ints are a bad comparison, since they have no structure anyway.  We also just derive `Debug` for `Complex` and `Rational`, leaving pretty formatting to `Display`.  Plus bigint's `to_str_radix` has to allocate memory and do a fair bit of computation, which I think is not a good idea for an error path.

If you have some have some other data structure which you want to make a pretty `Debug` for, you can always use `"{}"` for its `BigInt` components however you see fit.

ollie27:
I wasn't trying to make it prettier I was trying to make it more useful. I think decimal is more useful, do you think the current output is more useful?

Anyway it looks like `Display` is significantly slower than the derived `Debug` for really big numbers so for that reason I've removed the `Debug`  changes for now.

cuviper:
@homu r+ 95d8ccd

cuviper:
It depends -- if there was something wrong with bigint's implementation, then seeing the raw components would be more useful.  Otherwise, it is noisier, I do agree.  But since `Display`-like formatting is more costly, I prefer leaving it alone.

PR #161, opened 2016-01-28
Add signed/unsigned shifts to PrimInt

cuviper:


src/traits.rs
```diff
@@ -871,6 +939,22 @@ macro_rules! prim_int_impl {
                 <$T>::rotate_right(self, n)
             }
 
+            fn arithmetic_shl(self, n: u32) -> Self {
+                ((self as $S) << n) as $T
```
hauleth: Is it defined behaviour in Rust? It would be good to provide link to specs for that as I am not sure that it will be consistent between platforms.

cuviper: What exactly are you asking about?

Note that shl behaves the same either way, logical or arithmetic.  For shr, the [reference](http://doc.rust-lang.org/reference.html#bitwise-operators) calls it an arithmetic shift, but that doesn't actually do sign-extension for unsigned types -- much like C.

hauleth: Ok, so for signed it is defined as arithmetic shift, this should also be stated IMHO in documentation of `num`. Also it works incorrectly for unsigned integers bigger than `isize::max_integer()` as `0b1000_0000u8.arithmetic_shr(1)` would return `0b1100_0000u8` and I would rather expect true arithmetic shift right so it should be `0b0100_0000u8`.

cuviper: If we don't do "fake" sign extension, then there is no point to these new functions at all -- just use `Shr`.  Rather, I am treating it something like x86 `shar` which extends the most significant bit, as if it were signed, even though raw registers are not inherently signed/unsigned necessarily.

So I did document this -- "arithmetic" here is bitwise equivalent to signed shift, and "logical" is bitwise equivalent to unsigned shift.  The example explicitly shows the "sign"-extension even on u64, and doctest will verify this.

hauleth: That is the reason why I would change it's name to `signed` and `unsigned`. It will not confuse users.

cuviper: OK, I updated the names as you suggested.

hauleth:
Looks good to me.

@homu r+

PR #162, opened 2016-02-03
Add checked_pow function

est31:
Implements a `checked_pow` function which does the same as `pow`, just with overflow checks.

And, similar to #152 and #153, the function uses references instead of cloning.

Adds a little macro to spare code repetition. Its scoped to the function so nothing gets polluted.


cuviper:
Looks good -- it's nice that you mimicked `pow` so closely.
@homu r+ 3c5ecdc

Issue #102, opened 2015-06-08
Split num into several crates


vks:
Currently `num` mixes functionality that IMHO should be separate:

* bigints
* generic traits (`One`, `Zero`, `FromPrimitive`, `ToPrimitive`, ...)
* fractions
* complex numbers
* `range*`

The biggest offender are bigints, which are implemented very inefficiently. Real-world programs would probably rather use GMP or `ramp`.

The best plan of action would probably be creating the smaller crates and deprecating `num`?

bluss:
The traits are probably the most used part and still the most lightweight, they should definitely be their own crate.

vks:
I think we should create a new crate for the traits and the fractions (and maybe for `range*`) and abandon `num`. For complex numbers there is [stainless-steel/complex](https://github.com/stainless-steel/complex), for bigints there are [fizyk20/rust-gmp](https://github.com/fizyk20/rust-gmp) and [Aatch/ramp](https://github.com/Aatch/ramp). All of those are superior to the corresponding functionality in `num`.

cuviper:
I agree these would be better separate.  Really the only reason they're bundled is that `num` was a logical organization when they were all under the standard library, and they were pulled out en masse.  Maybe now that we have the rust-num organization, we can start splitting them into specialized repos, leaving `num` as an umbrella crate for compatibility.

In defense of bigint, I think it's nice to have a library that's stable, universal, and pure-rust.  Whereas rust-gmp depends on gmp, and ramp uses nightly features and arch-specific asm.  It's ok with me that those are better than bigint in other ways, perhaps especially performance.

kamalmarhubi:
I'd like to see the traits in their own crate as well.

vks:
To move this forward, we could extract the traits and publish a `num_traits` crate on crates.io.

vks:
This seems to be mostly done! The following function could still be moved to a sub crate crate: `checked_pow`, `pow`, `abs_sub`, `sub`, `one`, `zero`. Currently using them pulls in all crates, even if they are mostly not required.

bluss:
`one, zero` are basically redundant now; `A: Zero` allows calling `A::zero()` which it did not do sometime before Rust 1.0 when this was designed.

vks:
Still, I they could be moved to `num_traits`. Removing them would be a breaking change.

cuviper:
Sure, if you want to move them and re-export at the top, that should be
fine.

vks:
[...]

PR #164, opened 2016-02-15
Move segments of library to separate crates

hauleth:
Issue #102

- [x] traits
- [x] bigint
- [x] integer
- [x] complex
- [x] iter
- [x] rational


cuviper:
It's not *necessarily* a breaking change, if all of the same paths are still available through `num::foo`.

The travis scripts will need to be updated to make sure all tests are run across all subcrates.  Some of that is already extracted into separate shell scripts, so if it starts getting long we could pull the entire `script:` section into a real master script.

hauleth:
@cuviper I have changed some traits (i.e. content of `Num` trait was moved to the `FromStrRadix`) so it can brake some apps.

cuviper:
OK, I hadn't looked that closely since it's a WIP.  `FromStrRadix` sounds
like a reasonable but orthogonal change. Is it really necessary for the
split? Otherwise I'd do that separately.

hauleth:
No, it is not necessary. I will extract this. 

--
Łukasz Jan Niemier

Dnia 16 lut 2016 o godz. 19:18 Josh Stone <notifications@github.com> napisał(a):

kamalmarhubi:
I don't have a preference between this and my #174, except that it should happen sooner than later. Getting the traits into their own crate is already done, either here or there. If you get that merged separately, you can release the num-traits crate sooner. :-)

kamalmarhubi:
@hauleth can we get the traits change landed and released sooner? My thinking is that this won't be a breaking change for any users so there's no need to delay landing any of it.

kamalmarhubi:
@hauleth 
> @cuviper I have changed some traits (i.e. content of Num trait was moved to the FromStrRadix) so it can break some apps.

Well, aside from this being breaking. I think it's orthogonal to the crate refactor and should be done either before or after the multi-crate change.

hauleth:
I should finish this later today, so You can expect release today (UTC+1 timezone).

kamalmarhubi:
@hauleth ! yay!

hauleth:
Good to go.

cuviper:
Did you run through `rustfmt` or something?  That makes this a lot harder to compare... :-/

integer/src/lib.rs
```diff
@@ -255,18 +273,22 @@ macro_rules! impl_integer_for_isize {
             /// Calculates the Lowest Common Multiple (LCM) of the number and
             /// `other`.
             #[inline]
-            fn lcm(&self, other: &$T) -> $T {
+            fn lcm(&self, other: &Self) -> Self {
                 // should not have to recalculate abs
-                (*self * (*other / self.gcd(other))).abs()
+                ((*self * *other) / self.gcd(other)).abs()
```
cuviper: It appears #167 got lost here.

hauleth:
I have set `rustfmt` on save. 

--
Łukasz Jan Niemier

Dnia 11 mar 2016 o godz. 03:00 Josh Stone <notifications@github.com> napisał(a):

cuviper:
I really wish you hadn't.  That's only helpful for a project that already conforms, or maybe if you're the sole author, but here it has added a lot of noise for review.  I think I'll have to rustfmt master locally to get a clean basis for comparison, but I don't have time tonight.

Did you mean to claim authorship of the new crates, or was that just `cargo new` in action?  I think they should carry the same "Rust project" author as `num`, then add yourself individually if you want.

hauleth:
Ok, I will fix authors of crates. It was solely `cargo num` work (I haven't edited `Cargo.toml` manually as I am using `cargo-edit`).

cuviper:
I assume the remaining `src/bigint.rs` and `src/iter.rs` are just oversights?

hauleth:
Yes. 

--
Łukasz Jan Niemier

Dnia 12 mar 2016 o godz. 02:41 Josh Stone <notifications@github.com> napisał(a):

cuviper:
Looks like #176 also got lost.

cuviper:
OK, I've pored over this as closely as I can, and besides those things I commented it looks good.  Well, once the CI failures get fixed, of course.

I wonder if we can get access to do a [Crater](https://github.com/brson/taskcluster-crater) run... @brson?

cuviper:
Are you still actively working on this? Do you want me to add commits myself for the things I pointed out?

hauleth:
I will fix that right away. Sorry for my absence.

hauleth:
Due to unknown reasons (maybe bug in a compiler) code `num-traits` doesn't want to compile on 1.0.0

cuviper:
I found ways to workaround the 1.0.0 issues -- pushed.  I also fixed up the remaining review nits, and a few other build/test tweaks.  Let's see what travis thinks...

cuviper:
Success!

So, the final task is to update the descriptions in each subcrate `Cargo.toml` to something more specific.  Then when we finally publish all this, they should also get version numbers in lockstep with the main `num` crate.  If nothing else, that lets the signed git-tag make sense for all of them.

cuviper:
Aside, here's a neat way to test subpackages, pulling each `lib.rs` as a top-level `[[test]]`.
https://users.rust-lang.org/t/cargo-test-internal-packages/5187/3

hauleth:
@cuviper this hack is nice, but not as it would satisfy me. It requires that root crate will have the same dependencies as all sub-crates. This is kind of no-go for us as I've splitted dependencies. What is more I want to make `num-traits` `#![no_std]` which would make whole testing suite fail. So currently our only solution is to use `make`.

hauleth:
@cuviper are we ready with this?

cuviper:
We need better subcrate descriptions, but otherwise I think it's good.

bluss:
Will you bump the versions when you do this?

I have lots of ideas for Float that I'd like to explore if we have a new version. Ideally they would be in place in the first release of a new version.

cuviper:
No bump yet, I think semver should be OK through this one.
(Please speak up if you think there's a break!)

hauleth:
I will finish it tomorrow. What is left is descriptions and publishing all packages.

I think that we should tag each of the subcrates independently via `<subcrate name without num_* prefix>/<version>`.

cuviper:
@hauleth Have you worked on those descriptions?  If not, I can get to it tonight.

hauleth:
Here you are.

cuviper:
I'd prefer to keep a fuller description on the main `num` crate, but I'll tweak that myself.

I also just noticed that features aren't trickling down to subcrates correctly.  For example, enabling `serde` at the top doesn't enable it in `bigint`. (which is broken anyway, missing `extern crate serde;`)

hauleth:
@cuviper I know no solution to push flags to sub crates. Maybe there should be an RFC for that.

bluss:
I don't think there is a solution if the feature name is the same name as a dependency in the same crate. Otherwise you can forward like this:

```toml
[features]
serde = ["num-bigint/serde"]   # Requires same feature from dependency num-bigint
```

It is not conditionally forwarding.

This issue is semi related, isn't it? https://github.com/rust-lang/cargo/issues/1286

cuviper:
Ah, but the top-level `num` crate itself doesn't actually need any of those external dependencies (`rand`, `rustc-serialize`, `serde`).  So perhaps they can be left as abstract dependencies there, and then it should forward correctly, right?

Then for instance `rational` optionally uses `bigint`, but since the actual crate is named `num-bigint`, I think expanding `num-rational/bigint` may work.  We may need `default-features = false` in the top `Cargo.toml`'s rational section to allow fine control.

hauleth:
Should be fixed now.

cuviper:
Let's merge the beast.  I can try to publish tonight, if you don't get to it first.
May want to get #182 in there too - it's a simple regression fix.

@homu r+ 58b5fe5

cuviper:
This is all published now.

PR #168, opened 2016-02-21
Allow unary `+` when parsing Big{U,}Int from strings

crazymykl:


cuviper:
I think this will do the wrong thing for a double-sign, e.g. `"-+10".parse()` should be an error.
Please add a test for that and fix it. :)

For consistency, it would be nice to implement this for `BigUint` too.

src/bigint.rs
```diff
@@ -366,6 +366,16 @@ impl Num for BigUint {
     /// Creates and initializes a `BigUint`.
     fn from_str_radix(s: &str, radix: u32) -> Result<BigUint, ParseBigIntError> {
         assert!(2 <= radix && radix <= 36, "The radix must be within 2...36");
+        let mut s = s;
+        if s.starts_with('+') {
+            s = &s[1..];
+            if s.starts_with('+') {
+                // create ParseIntError::InvalidDigit
```
crazymykl: I really don't like this pattern, but it's what's used elsewhere.

crazymykl:
I'm not thrilled with the error-handling code, but it's consistent with what's there.

cuviper:
Yeah, the error handling trying to wrap `ParseIntError` is clumsy.  It was probably not as bad when all this was together in the standard library.  But `ParseBigIntError` is a `pub enum`, so we'll have to treat that as a breaking change to alter it.

For your part, I think you can dodge dealing with it directly though.  I'll comment inline.

src/bigint.rs
cuviper: Must you handle a second '+' specifically though?  Seems if you just let it continue, that would get flagged invalid in the byte-conversion step.  I guess you avoid creating that Vec, but there's no need to optimize for error cases.

You should add a "++" test either way, since this is a case you thought about.

crazymykl: I did create a test for it, see line 3767. The reason I need to check for this separately is because the error path in the byte-conversion step is `Ok` with `"+21"`.

src/bigint.rs
```diff
@@ -1891,7 +1901,15 @@ impl Num for BigInt {
...
     #[inline]
     fn from_str_radix(mut s: &str, radix: u32) -> Result<BigInt, ParseBigIntError> {
-        let sign = if s.starts_with('-') { s = &s[1..]; Minus } else { Plus };
+        let sign = if s.starts_with('-') {
+            s = &s[1..];
+            if s.starts_with('+') {
+                // create ParseIntError::InvalidDigit
+                let e = i64::from_str_radix("-+1", radix).unwrap_err();
+                return Err(e.into());
+            }
```
cuviper: What if instead, you only take the substring consuming '-' when it's not followed by '+'.  So when there is a "-+", it just passes the whole thing to BigUint, which will barf on '-'.

    if s.starts_with('-') {
        // Only consume the minus sign if it's alone
        let tail = &s[1..];
        if !tail.starts_with('+') { s = tail; }
    ...

src/bigint.rs
cuviper: Ah indeed, sorry, I missed that test.  Maybe try the same thing I just suggested for "-+"?
i.e. don't commit the substring if there's a double-sign.

crazymykl: This actually fixes a panic in the current code:

```rust
extern crate num;

fn main() {
    println!("{:?}", "+1".parse::<num::BigUint>());
}
```

src/bigint.rs
crazymykl: I like this a lot better than creating a synthetic failure.

crazymykl:
That's way cleaner, thanks.

cuviper:
Thanks!

@homu r+ 9def836

PR #169, opened 2016-02-23
Avoid overflows in Ratio's Ord::cmp

### LINKED ISSUE #7: Comparison of `num::rational::Ratio` is incorrect with overflow
<a href="https://github.com/huonw"></img></a> **Issue by [huonw](https://github.com/huonw)**
_Monday Aug 12, 2013 at 09:23 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/8466_

_This issue was labelled with: A-libs in the Rust repository_

----

```rust
extern crate num;
use num::rational::Ratio;

fn main() {
    let big = Ratio::new(128u8, 1);
    let small = big.recip();
    assert!(big > small);
}
```

fails the assertion. (It's comparing `128` with `1/128`.) 

This, or something similar, happens for any type that can overflow (I guess one could argue that fixed size types are "unsupported" for rational numbers, and so this overflow leads to unspecified behaviour).

cuviper:
Fixes #7

### DIFF
--- a/src/rational.rs
+++ b/src/rational.rs
@@ -224,33 +224,67 @@ impl Ratio<BigInt> {
 
 /* Comparisons */
 
-// comparing a/b and c/d is the same as comparing a*d and b*c, so we
-// abstract that pattern. The following macro takes a trait and either
-// a comma-separated list of "method name -> return value" or just
-// "method name" (return value is bool in that case)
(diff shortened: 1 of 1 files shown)


hauleth:
@homu r+

PR #174, opened 2016-03-03
Move traits into num-traits crate

kamalmarhubi:
This commit moves the traits module into its own crate. The new crate is
publicly imported so that downstream code will work without changes. New
code can choose to only depend on the traits crate.

Refs https://github.com/rust-num/num/issues/102


kamalmarhubi:
It should be possible for this to be a minor release.

cuviper:
See also #164 -- you and @hauleth will have to decide how to coordinate on this.

kamalmarhubi:
@cuviper oh I didn't realize there was already a PR open to do it.

kamalmarhubi:
Looks like #164 is landing soon, so I'll drop this.

PR #175, opened 2016-03-04
Saturating<T> wrapper type

m4rw3r:
Rust's stdlib has a [`Wrapping<T>`](http://doc.rust-lang.org/nightly/std/num/struct.Wrapping.html) wrapper type for primitive integer types. The type provides intentionally wrapping semantics to all operations performed on it.

This pull-request introduces a `Saturating<T>` wrapper type which is the corresponding saturating wrapper type for primitive integer types. This type provides saturating semantics on arithmetic and bit-shifting operations for both signed and unsigned integers.

The goal is to provide a type which is easy to slot in instead of either `Wrapping<T>` or any basic integer type. Combined with the traits in `num` it would be possible to make functions/methods generic over the type of integer and if it should have wrapping or saturating semantics.

The name `Saturating<T>` collides with the trait `Saturating` which is unfortunate. I suspect that this type will need a name change but at the moment I do not have any suggestions.

### DIFF
--- a/src/saturating.rs
+++ b/src/saturating.rs
@@ -31,7 +31,7 @@ use traits::{Bounded, Num, One, Signed, Unsigned, Zero};
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Default)]
 pub struct Saturating<T>(pub T);
 
-macro_rules! impl_saturating {
+macro_rules! impl_saturating_base {
     ( $t:ty ) => {
         impl Add for Saturating<$t> {
             type Output = Saturating<$t>;
@@ -131,8 +131,6 @@ macro_rules! impl_saturating {
 
 macro_rules! impl_saturating_unsigned {
     ( $t:ty ) => {
-        impl_saturating!{$t}
-
         impl Unsigned for Saturating<$t> {}
 
         impl Mul for Saturating<$t> {
@@ -164,13 +162,10 @@ macro_rules! impl_saturating_unsigned {
             }
         }
     };
-    ( $($t:ty)* ) => { $(impl_saturating_unsigned!{$t})* };
 }
 
 macro_rules! impl_saturating_signed {
     ( $t:ty ) => {
-        impl_saturating!{$t}
-
         impl Mul for Saturating<$t> {
             type Output = Saturating<$t>;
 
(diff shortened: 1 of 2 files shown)


src/saturating.rs
```diff
@@ -0,0 +1,619 @@
...
+
+        impl Bounded for Saturating<$t> {
+            #[inline(always)]
+            fn min_value() -> Self {
+                Saturating(<$t>::min_value())
+            }
+
+            #[inline(always)]
+            fn max_value() -> Self {
+                Saturating(<$t>::min_value())
```
SoniEx2: Uh this doesn't look right...

m4rw3r: Oops

cuviper:
`Saturating<T>` was also just suggested for the standard library in https://github.com/rust-lang/rfcs/pull/1218#issuecomment-192354182.  It's a reasonable idea, but don't you think it would make more sense implemented next to `Wrapping<T>`?

I'm not sure that I agree that implementing `num` traits for this is helpful.  Notice we have nothing yet implemented for `Wrapping<T>` either.  I feel like it would be surprising for a generic function taking `Num` to encounter wrapping or saturating semantics.  Can you give a concrete example where this would really be useful to you?

m4rw3r:
@cuviper I guess it depends on the number of people who think it would be useful, but it does make more sense to provide both `Wrapping<T>`, `Saturating<T>` and maybe even some `Checked<T>` (always panic on overflow) in the standard library.

As for usefulness in the generic case I have some code which is read from external input where I do not mind if the values are saturated if they are out of bounds, here it is useful with a saturating datatype since I do not have to manually perform checks to make sure that the data I read is within the range of the datatype.

SoniEx2:
Would be nice to have traits for wrapping_add(), saturating_add(), etc, and have a Wrapping<T>, Saturating<T>, etc impl for them. But that's another issue.

cuviper:
@m4rw3r then will you propose this for the standard library instead?
If that's rejected, then we can reconsider it here...

m4rw3r:
@cuviper Submitted as an RFC here: https://github.com/rust-lang/rfcs/pull/1534  this also includes the always-checked counterpart too.

cuviper:
If this is ever revived, it will need to be on the separate repo:
https://github.com/rust-num/num-traits

PR #180, opened 2016-03-07
Fix ToPrimitive for f64 -> f32 conversion.

bluss:
Fix ToPrimitive for f64 -> f32 conversion.

It should use the destination type and not the source type to check if
the conversion would be to a value that's in range.

NOTE: A finite f64 value that is larger than the f32 value range now produces
None when converted to f32 with ToPrimitive.

Previously, too large f64 values would produce inf f32 values. This `as`
cast has an undefined result and was not specified to always produce for
example `inf`.

The conversion preserves nan/+-inf specifically.

### DIFF
--- a/traits/src/cast.rs
+++ b/traits/src/cast.rs
@@ -222,9 +222,11 @@ macro_rules! impl_to_primitive_float_to_float {
         if size_of::<$SrcT>() <= size_of::<$DstT>() {
             Some($slf as $DstT)
         } else {
+            // Make sure the value is in range for the cast.
+            // NaN and +-inf are cast as they are.
             let n = $slf as f64;
-            let max_value: $SrcT = ::std::$SrcT::MAX;
-            if -max_value as f64 <= n && n <= max_value as f64 {
+            let max_value: $DstT = ::std::$DstT::MAX;
+            if !n.is_finite() || (-max_value as f64 <= n && n <= max_value as f64) {
                 Some($slf as $DstT)
             } else {
                 None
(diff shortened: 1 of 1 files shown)


bluss:
I guess this is stopping because of the ongoing trait refactoring? That's ok, I think splitting is a good idea.

cuviper:
Yeah, sorry I didn't say so. I'm holding off on approving more PRs until
that goes through.

bluss:
Updated. This is a breaking change but also a bug fix.

bluss:
The effect of the bug was that it always return Some for floats. Maybe that's a bad thing to change without a breaking change? The current code is just nonsensical and we could change it into just the cast without the `if`.

bluss:
Hm ok it's ~~always~~not always. inf is always converted to None.

cuviper:
I think it's safe to classify this as a pure bugfix, not a breaking change, as the intent was clearly there to have a range check.  Otherwise, we'll never be able to change anything! (ref https://xkcd.com/1172/)

However, there are some strange edge cases here.  If I'm reading correctly, f32->f64 inf will succeed, but either f32->f32 or f64->f64 will be None.  I think it's reasonable to expect inf and NaN to convert all around, but then shouldn't large f64 just be f32 inf anyway?

cuviper:
(nevermind, comparing sizes `<=` makes self-conversion stable.  I read it as just `<`)

bluss:
I had to do research. Now we have even better backing for changing the behavior. Casting a too large f64 to f32 is UB according to [langref](http://llvm.org/docs/LangRef.html#fptrunc-to-instruction). This is indeed the instruction rust uses [playground link](https://play.rust-lang.org/?gist=dd7184b249cabe3e442b8afe84a54d4d&version=nightly&backtrace=0) (Nightly/Release/LLVM IR to see).

Now this isn't fully num's concern to solve, core Rust will have to solve this and I think it's pretty well known (the trouble with `as` and floating point casts). But it means that the range check in the method is well intentioned.

cuviper:
Can you make it so +/-inf and NaN carry over from f64 to f32?

The range check *could* decide to manually create inf for out of range values, dodging the undefined result.  But maybe None is better so there's a signal that it didn't really convert.

(And then float->int is completely unchecked... #135 was working on that.)

bluss:
Yes, I will come back to it tomorrow.

bluss:
Updated commit & PR message. It now preserves nan/inf (using f32::is_finite())

cuviper:
Thanks!

@homu r+ acde249

PR #182, opened 2016-03-10
Correct Ratio::is_negative and Ratio::is_positive

ollie27:
Zero is not positive or negative.

This was broken in 8be7e7bab5ffc7f998e713547b67c1a424d3b9b1.

### DIFF
--- a/rational/src/lib.rs
+++ b/rational/src/lib.rs
@@ -526,12 +526,14 @@ impl<T: Clone + Integer + Signed> Signed for Ratio<T> {
 
     #[inline]
     fn is_positive(&self) -> bool {
-        !self.is_negative()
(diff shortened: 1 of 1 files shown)


cuviper:
Ah, good catch!  Looks right -- I just want to hold this for a bit so #164 isn't disrupted...

ollie27:
Rebased.

cuviper:
Just noticed, is there a reason you used bitwise `&` and `|` instead of lazy-boolean `&&` and `||`?

ollie27:
I've no idea, I've changed it to `&&` and `||` anyway.

cuviper:
Thanks!

@homu r+ c22e3bf

PR #188, opened 2016-04-20
bigint: Fix calculation of size for multiply temporaries

### LINKED ISSUE #187: Failing assertion while multiplying two BigUints
```
extern crate num;
use num::CheckedMul;
use num::bigint::BigUint;
 
fn main() {
    let s = "531137992816767098689588206552468627329593117727031923199444138200403559860852242739162502232636710047537552105951370000796528760829212940754539968588340162273730474622005920097370111";
    let a: BigUint = s.parse().unwrap();
    let b = a.clone();
    let _ = a.checked_mul(&b);
}
```
The code above crashes at [this assertion in num-bigint](https://github.com/rust-num/num/blob/master/bigint/src/lib.rs#L783):
```
thread '<main>' panicked at 'assertion failed: carry == 0', /home/virgile/.cargo/registry/src/github.com-88ac128001ac3a9a/num-bigint-0.1.32/src/lib.rs:783
```
while it runs fine with one digit less in `s`. The same problem occurs with an unchecked multiplication.

I am not sure how to interpret the comment below the assertion; is there an allocation I should make manually?

koverstreet:
When x.len() and y.len() are both equal and odd, we have x1.len() + y1.len() + 1
(the required size to multiply x1 and y1) > y.len() + 1

This fixes #187

### DIFF
--- a/bigint/src/lib.rs
+++ b/bigint/src/lib.rs
@@ -991,9 +991,11 @@ fn mac3(acc: &mut [BigDigit], b: &[BigDigit], c: &[BigDigit]) {
         let (x0, x1) = x.split_at(b);
         let (y0, y1) = y.split_at(b);
 
-        /* We reuse the same BigUint for all the intermediate multiplies: */
-
-        let len = y.len() + 1;
+        /*
+         * We reuse the same BigUint for all the intermediate multiplies and have to size p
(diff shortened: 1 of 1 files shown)


cuviper:
Great! Can you add the test that failed #187?

koverstreet:
Done

cuviper:
Thanks a lot!

@homu r+ 1e65e9d

PR #190, opened 2016-04-29
num: remove `pub` from sub-crates

### LINKED ISSUE #189: warning: `pub extern crate` does not work as expected and should not be used
Instead of, e.g. `pub extern crate num_traits`, the code should use `extern crate num_traits; pub use num_traits;`

Otherwise, rustc causes warnings to be output for any crate that uses `num`.

cuviper:
Rust 1.4 through 1.7 will issue warnings that `pub extern crate` does
not work as expected, although it appears to be fine.  But Rust 1.8 and
later issue `#[warn(private_in_public)]` for `pub use num_foo as foo` if
that crate isn't public, and that's headed toward a hard error.

@bluss suggested instead `pub mod foo { pub use num_foo::*; }`, which
I thought I had tried before, but it appears to work fine on all
versions of Rust.  Let's do it!

A small downside is that docs for `num::foo` now just show the wildcard
reexport, instead of direct documentation, but at least there's a link
to follow to the sub-crate.

It's a breaking change that `num::num_foo` paths are no longer public,
but we didn't really want those exposed in the first place.  I consider
this minor -- people should either use the `num::foo` module as before
the split-up, or use the `num_foo` crate directly.

Fixes #189.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,15 +57,15 @@
        html_root_url = "http://rust-num.github.io/num/",
        html_playground_url = "http://play.rust-lang.org/")]
 
-pub extern crate num_traits;
-pub extern crate num_integer;
-pub extern crate num_iter;
+extern crate num_traits;
+extern crate num_integer;
+extern crate num_iter;
 #[cfg(feature = "num-complex")]
-pub extern crate num_complex;
+extern crate num_complex;
 #[cfg(feature = "num-bigint")]
-pub extern crate num_bigint;
+extern crate num_bigint;
 #[cfg(feature = "num-rational")]
-pub extern crate num_rational;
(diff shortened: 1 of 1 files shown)


bluss:
I only mentioned it exists, don't know if it will work well.

cuviper:
That's partly why I opened a PR instead of pushing this right away.  It looks fine so far.

cuviper:
@homu r+

hauleth:
In Octavo I am using `pub extern crate name as other_name;` which has that advantage that this provide better documentation.

cuviper:
Sure, that would be nice, but it runs afoul of the same warning as #189 in older rustc.

PR #199, opened 2016-06-29
Implement Default for Complex

### LINKED ISSUE #198: Complex should implement Default
Title says it all.

adamcrume:
Fixes #198


cuviper:
Can't it just be added to the `derive` attribute?

adamcrume:
I thought that would prevent you from creating a Complex<T> for a T that doesn't implement Default, but apparently that's not the case.  I changed it to use derive.

hauleth:
@homu r+

Issue #202, opened 2016-06-30
Implement more number theory functions


ejmahler:
In num_integer, I see an implementation of gcd.

I would like to implement some related number theory functions: [modular exponentiation](https://en.wikipedia.org/wiki/Modular_exponentiation), [extended euclid's algorithm](https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm), [modular multiplicative inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse).

I'm also interested in prime factorization, euler's totient function, primitive roots, etc, but these might be too much.

Is there an interest for these functions in this library, to go alongside gcd? I'd be happy to implement the functions, tests, documentation, etc, I just want to make sure there's interest from the maintainers before I begin.

hauleth:
It would be great!

koverstreet:
I'd like to see these implemented as generics over ring/field traits. Especially with specialization coming soon (or has it landed?), this is really the right way to go.

ejmahler:
Could you give a really quick example, to illustrate what you're picturing?

koverstreet:
Along the lines of a numeric tower (read up on Haskell's if you're curious) - except, I have no intention of trying to do a full numeric tower because no one seems to be able to get that right.

All I'm thinking for now is - define traits for Group, Ring, Field, etc. so we have a place to stick generic algorithms. I tend to think we shouldn't even have the various traits inherit from each other, at least for now - BigUint can always inherit from each of Group/Ring/whatever.

Basically keep it as simple as possible for now - nothing more than a place to stick things that should work on anything that's a group/ring/field.

gcd(), lcm()... perhaps a few others should then become generic functions on whatever the appropriate trait is, with specializations where it matters for performance.

hauleth:
I think that having additional structures defined using typenum would feel more "natural".

ejmahler:
I understand the high-level picture of what you're looking for, @koverstreet . A set of composable traits that can be included and maybe specialized for each type.

So starting small for now with jsut a "ring" trait that implements all sorts of modular arithmetic functions on this -- what's the correct way to implement this? I wouldn't call myself a rust expert yet so I need some guidance.

The following is what my mind jumps to, but this obviously doesn't compile because you can't implement a trait in terms of another trait -- you need a concrete type. But if we implement this trait for all the primitive types explicitly, that defeats the point of composition of these traits, doesn't it?

```
pub trait Ring {
[...]

PR #203, opened 2016-07-05
Add two new number theory functions to the Integer trait

ejmahler:
Made as a part of issue #202 

This PR adds two methods to the Integer trait. The first is modular exponentiation (`pow_mod`). The second is multiplicative inverse modulo (`inverse_mod`).

I'd appreciate some scrutiny on my big int implementations of these functions. Although they pass all the tests, I'm sure there are optimizations to be made to avoid unnecessarily making copies of numbers, etc.

### DIFF
--- a/bigint/src/lib.rs
+++ b/bigint/src/lib.rs
@@ -4008,7 +4008,7 @@ mod biguint_tests {
             assert_eq!(big_base.inverse_mod(&big_modulo), None);
         }
 
-        // if the modulo is prome we're guaranteed to get an answer
+        // if the modulo is prime we're guaranteed to get an answer
         check(15, 17, 8);
 
         // gcd(4, 10) = 2, so 4 doesn't have an inverse mod 10
@@ -4016,6 +4016,12 @@ mod biguint_tests {
 
         // but gcd(3, 10) = 1, so 3 has an inverse mod 10
         check(3, 10, 7);
+
+        //0 does not have an inverse
+        check_none(0, 10);
+
+        //the inverse of 1 is 1
+        check(1, 10, 1);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


bigint/src/lib.rs
```diff
@@ -1263,6 +1263,65 @@ impl Integer for BigUint {
...
+            One::one()
+        } else {
+            base = base % modulo;
+
+            while !exp.is_zero() {
+                if exp.is_odd() {
+                    result = result * &base % modulo;
+                }
+                exp = exp >> One::one();
+                base = &base * &base % modulo;
```
vks: You probably could avoid a copy by taking one `base` by value.

bigint/src/lib.rs
```diff
@@ -2401,6 +2460,79 @@ impl Integer for BigInt {
...
```
vks: One `base` could be taken by value.

vks:
There seems to be quite a bit of code duplication with regards to Euclid's algorithm to calculate the GCD. It would be nice to have that as a separate function, because it can be useful on its own. Also it would make it easier to replace with a more efficient binary algorithm.

ejmahler:
Unfortunately, the GCD calculation is inseparable from the multiplicative inverse calculation. We're not merely calculating the GCD in order to verify that an inverse exists, the inverse calculation is an "addition" to the euclidean algorithm: https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm

vks:
You can implement the extended Euclidean algorithm as a separate function. See [my take](https://github.com/vks/discrete-log/blob/master/src/main.rs) on implementing the modular power and inverse.

ejmahler:
Definitely - I decided not to so that we aren't doing the extra work of computing the other coefficient. If you think it's a better idea to compute it and ignore it in the name of cleaner code I'd be happy to do that.

If we go that route, how should we handle unsigned numbers with the extended euclidean algorithm? It commonly returns negative numbers. Should we just not implement it for unsigned numbers?

vks:
You could also have a function that does not calculate the second coefficient. If you want to go full DRY, you could use a macro and dead code elimination.

I think for unsigned numbers Euclid's extended algorithm only makes sense if you implent it for modular numbers only.

ejmahler:
If we're not implementing it for unsigned types, then it obviously can't be a method on the Integer trait. Where do you think it should go instead?

cuviper:
I would definitely like to include unsigned types for `pow_mod` and `inverse_mod`, even if that means it can't use a "pure" version of extended GCD.

ejmahler:
Closing because I'm getting the vibe this isn't acceptable without pretty big changes

Issue #151, opened 2016-01-06
The dependencies of the `bigint` feature on `rand` and `rustc_serialize` should be optional.


briansmith:
My builds are failing on Nightly because I depend on `num` and the `rand` crate fails to build on Nightly. My crate only depends on `rand` because it depends on `BigInt`. I tried to build without the `rand` dependency but that made `BigInt` unavailable.

Instead all of `BigInt` being dependent on the `rand` feature, only the random-related parts of `Bigint` should depend on the `rand` feature. And, in particular, when building `num` without the `rand` feature, the rest of `BigInt` should work.

briansmith:
See https://github.com/rust-lang/rust/issues/30713.

briansmith:
Oops, here's the real issue: If you build with the `bigint` feature (which is the default), then there's a non-optional dependency on `rand` and `rustc_serialize`. There should be some way to make that dependency optional and for users of `num` to build without those dependencies.

cuviper:
Sure, making these fully optional would be welcome.

It used to be mandatory for `bigint` to have `rust-serialize` for `ToHex`, but I nixed that.  The remaining derives could be configured like they are in `complex` and `rational`.

So these two would be dropped from `bigint` dependencies, then added to default feature list instead.

briansmith:
Just to be clear, you would accept this?:

* Add two new features: `serialize` and `rand`.
* Add those two features to the default feature set.
* Make the `FromHex`/`ToHex` parts of `bigint` conditional on the `serialize` feature.
* Make the `rand`-dependent parts of `bigint` conditional on the `rand` feature.
* Remove the dependencies on `rustc_serialize` and `rand` from the `bigint` feature.

briansmith:
My pull request #154 fixes this.

> Add two new features: serialize and rand.

Optional dependencies are handled like features, so I didn't need to do this.

> Add those two features to the default feature set.

I did this.

> Make the FromHex/ToHex parts of bigint conditional on the serialize feature.

I made it conditional on the "rustc_serialize" feature.

> Make the rand-dependent parts of bigint conditional on the rand feature.

I did this, but I also had to enable the `rand` functionality for all test builds because some of the tests are randomized using the `rand` functionality.

> Remove the dependencies on rustc_serialize and rand from the bigint feature.

I did this.

cuviper:
#154 is merged.

briansmith:
[...]

PR #204, opened 2016-07-08
Minor optimization, prep work for more optimization

koverstreet:
The patch "drop some dependencies on BigDigit's size" is the one I'd really like to get in.

### DIFF
--- a/bigint/src/lib.rs
+++ b/bigint/src/lib.rs
@@ -306,7 +306,8 @@ impl FromStr for BigUint {
     }
 }
 
-// Read bitwise digits that evenly divide BigDigit
+// Convert from a power of two radix (bits == ilog2(radix)) where bits evenly divides
+// BigDigit::BITS
 fn from_bitwise_digits_le(v: &[u8], bits: usize) -> BigUint {
     debug_assert!(!v.is_empty() && bits <= 8 && big_digit::BITS % bits == 0);
     debug_assert!(v.iter().all(|&c| (c as BigDigit) < (1 << bits)));
@@ -315,14 +316,15 @@ fn from_bitwise_digits_le(v: &[u8], bits: usize) -> BigUint {
 
     let data = v.chunks(digits_per_big_digit)
                 .map(|chunk| {
-                    chunk.iter().rev().fold(0u32, |acc, &c| (acc << bits) | c as BigDigit)
+                    chunk.iter().rev().fold(0, |acc, &c| (acc << bits) | c as BigDigit)
                 })
                 .collect();
 
     BigUint::new(data)
 }
 
-// Read bitwise digits that don't evenly divide BigDigit
+// Convert from a power of two radix (bits == ilog2(radix)) where bits doesn't evenly divide
+// BigDigit::BITS
 fn from_inexact_bitwise_digits_le(v: &[u8], bits: usize) -> BigUint {
     debug_assert!(!v.is_empty() && bits <= 8 && big_digit::BITS % bits != 0);
     debug_assert!(v.iter().all(|&c| (c as BigDigit) < (1 << bits)));
@@ -331,15 +333,20 @@ fn from_inexact_bitwise_digits_le(v: &[u8], bits: usize) -> BigUint {
     let mut data = Vec::with_capacity(big_digits);
 
     let mut d = 0;
-    let mut dbits = 0;
+    let mut dbits = 0; // number of bits we currently have in d
+
+    // walk v accumululating bits in d; whenever we accumulate big_digit::BITS in d, spit out a
+    // big_digit:
     for &c in v {
-        d |= (c as DoubleBigDigit) << dbits;
+        d |= (c as BigDigit) << dbits;
         dbits += bits;
+
         if dbits >= big_digit::BITS {
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            data.push(lo);
-            d = hi as DoubleBigDigit;
+            data.push(d);
             dbits -= big_digit::BITS;
+            // if dbits was > big_digit::BITS, we dropped some of the bits in c (they couldn't fit
+            // in d) - grab the bits we lost here:
+            d = (c as BigDigit) >> (bits - dbits);
         }
     }
 
@@ -362,8 +369,7 @@ fn from_radix_digits_be(v: &[u8], radix: u32) -> BigUint {
     let mut data = Vec::with_capacity(big_digits as usize);
 
     let (base, power) = get_radix_base(radix);
-    debug_assert!(base < (1 << 32));
-    let base = base as BigDigit;
+    let radix = radix as BigDigit;
 
     let r = v.len() % power;
     let i = if r == 0 {
@@ -435,7 +441,7 @@ impl Num for BigUint {
 
         let res = if radix.is_power_of_two() {
             // Powers of two can use bitwise masks and shifting instead of multiplication
-            let bits = radix.trailing_zeros() as usize;
+            let bits = ilog2(radix);
             v.reverse();
             if big_digit::BITS % bits == 0 {
                 from_bitwise_digits_le(&v, bits)
@@ -1349,6 +1355,46 @@ impl Integer for BigUint {
     }
 }
 
+fn high_bits_to_u64(v: &BigUint) -> u64 {
+    match v.data.len() {
+        0   => 0,
+        1   => v.data[0] as u64,
+        _   => {
+            let mut bits = v.bits();
+            let mut ret = 0u64;
+            let mut ret_bits = 0;
+
+            for d in v.data.iter().rev() {
+                let digit_bits = (bits - 1) % big_digit::BITS + 1;
+                let bits_want = cmp::min(64 - ret_bits, digit_bits);
+
+                if bits_want != 64 {
+                    ret <<= bits_want;
+                }
+                ret      |= *d as u64 >> (digit_bits - bits_want);
+                ret_bits += bits_want;
+                bits     -= bits_want;
+
+                if ret_bits == 64 {
+                    break;
+                }
+            }
+
+            ret
+        }
+    }
+}
+
+/// Find last set bit
+/// fls(0) == 0, fls(u32::MAX) == 32
+fn fls<T: traits::PrimInt>(v: T) -> usize {
+    std::mem::size_of::<T>() * 8 - v.leading_zeros() as usize
+}
+
+fn ilog2<T: traits::PrimInt>(v: T) -> usize {
+    fls(v) - 1
+}
+
 impl ToPrimitive for BigUint {
     #[inline]
     fn to_i64(&self) -> Option<i64> {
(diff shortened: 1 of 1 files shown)


cuviper:
I'd prefer to see these as separate PRs, so we can keep focus separately.  It may be that some parts are too intertwined to work separately, but so be it.  Let's do it incrementally.  And CI needs to pass, of course -- you have some rust-1.0 issues.

- `Mul` with primitives is good, but I'll want the full complement.  All ops, primitive either on left or right, and all variations of val/ref (including primitives by reference).  More macros will probably help.

- Shifting -- what's the performance like without `unsafe`?  We have very minimal use of `unsafe` so far, all easily verified, and I'd really rather not open the door to big blocks of code like this.

- Starting to isolating `BigDigit` is fine, but I'm going to be really hesitant to make bigger changes until we jump more fully into #151.  These changes should show performance comparisons too, please.

koverstreet:
Sounds good - this PR is now just the BigDigit changes, leaving the others for now. Added benchmark results.

bigint/src/lib.rs
```diff
@@ -127,6 +127,7 @@ pub mod big_digit {
 
     // `DoubleBigDigit` size dependent
     pub const BITS: usize = 32;
+    pub use ::std::u32::MAX;
```
cuviper: Let's not add any new `pub` items please.  In a perfect world, I wouldn't even make `big_digit` public at all...
Just make it a top-level private name if you need to use it throughout the module.

bigint/src/lib.rs
```diff
@@ -333,13 +336,13 @@ fn from_inexact_bitwise_digits_le(v: &[u8], bits: usize) -> BigUint {
...
+        d |= (c as BigDigit) << dbits;
         dbits += bits;
+
         if dbits >= big_digit::BITS {
-            let (hi, lo) = big_digit::from_doublebigdigit(d);
-            data.push(lo);
-            d = hi as DoubleBigDigit;
+            data.push(d);
             dbits -= big_digit::BITS;
+            d = (c as BigDigit) >> (bits - dbits);
```
cuviper: At first I didn't see how the `<< dbits` above could work without overflowing, but then I figured out that you're recovering those lost bits here.  This deserves some quick comments to explain.

bigint/src/lib.rs
```diff
@@ -1349,6 +1351,40 @@ impl Integer for BigUint {
...
+    match v.data.len() {
+        0   => 0,
+        1   => v.data[0] as u64,
+        _   => {
+            let mut bits = v.bits();
+            let mut ret = 0u64;
+            let mut ret_bits = 0;
+
+            for d in v.data.iter().rev() {
+                let digit_bits = ((bits - 1) & (big_digit::BITS - 1)) + 1;
```
cuviper: AKA `bits % BITS`, right?  Just write that, and let the optimizer be the clever one.

bigint/src/lib.rs
```diff
@@ -1349,6 +1351,40 @@ impl Integer for BigUint {
...
+            }
+
+            ret
+        }
+    }
+}
+
+fn ilog2(v: u64) -> usize {
+    (64 - v.leading_zeros()) as usize
+}
```
cuviper: This `ilog2` isn't really mathematically correct, e.g. for 4 it will return (64-61)=3.  So it's more like `floor(log2(v))+1`, or in programming terms it's simply the number of significant bits.  Let's rename it to avoid confusion, please.

cuviper: It's basically the same as `BigUint::bits`, so maybe call it `u64_bits`.

bigint/src/lib.rs
```diff
@@ -1484,14 +1497,16 @@ impl FromPrimitive for BigUint {
...
-        match big_digit::from_doublebigdigit(n) {
-            (0, 0) => BigUint::zero(),
-            (0, n0) => BigUint { data: vec![n0] },
-            (n1, n0) => BigUint { data: vec![n0, n1] },
+    fn from(mut n: u64) -> Self {
+        let mut ret: BigUint = Zero::zero();
+
+        while n != 0 {
+            ret.data.push(n as BigDigit);
+            n = (n >> 1) >> (big_digit::BITS - 1);
```
cuviper: I understand you're avoiding shift overflow if `BITS` were 64, but this needs a comment.

bigint/src/lib.rs
```diff
@@ -1591,28 +1606,35 @@ fn to_bitwise_digits_le(u: &BigUint, bits: usize) -> Vec<u8> {
...
+    let mask: BigDigit = (1 << bits) - 1;
     let digits = (u.bits() + bits - 1) / bits;
     let mut res = Vec::with_capacity(digits);
 
     let mut r = 0;
     let mut rbits = 0;
-    for hi in u.data[..last_i].iter().cloned() {
-        r |= (hi as DoubleBigDigit) << rbits;
+
+    for c in &u.data[..] {
```
cuviper: The `[..]` is unnecessary, just `for c in &u.data` will do.

bigint/src/lib.rs
```diff
@@ -1591,28 +1606,35 @@ fn to_bitwise_digits_le(u: &BigUint, bits: usize) -> Vec<u8> {
...
+        r |= *c << rbits;
         rbits += big_digit::BITS;
 
         while rbits >= bits {
             res.push((r & mask) as u8);
             r >>= bits;
+
+            if rbits > big_digit::BITS {
+                r = *c >> (big_digit::BITS - (rbits - bits));
```
cuviper: Add another "bit-recovery" note, please.

bigint/src/lib.rs
```diff
@@ -1852,57 +1873,115 @@ impl serde::Deserialize for BigUint {
...
+                (387420489,				6), // 27
+                (481890304,				6), // 28
+                (594823321,				6), // 29
+                (729000000,				6), // 30
+                (887503681,				6), // 31
+                (0,						0), // 32
+                (1291467969,			6), // 33
+                (1544804416,			6), // 34
+                (1838265625,			6), // 35
+                (2176782336,			6) /* 36 */
```
cuviper: Please use spaces inside this table for alignment, rather than tabs, and make that final 36 comment consistent.

bigint/src/lib.rs
```diff
@@ -3469,23 +3547,20 @@ mod biguint_tests {
 
         check(BigUint::new(vec![]), 0);
         check(BigUint::new(vec![1]), (1 << (0 * big_digit::BITS)));
-        check(BigUint::new(vec![N1]), (1 << (1 * big_digit::BITS)) - 1);
-        check(BigUint::new(vec![0, 1]), (1 << (1 * big_digit::BITS)));
-        check(BigUint::new(vec![N1, N1 >> 1]), i64::MAX);
+        check((BigUint::one() << 63) - BigUint::one(), i64::MAX);
```
cuviper: These tests shouldn't be changed.  Whatever people might give on the public interfaces needs to keep working!

bigint/src/lib.rs
```diff
@@ -3495,9 +3570,7 @@ mod biguint_tests {
 
         check(BigUint::new(vec![]), 0);
         check(BigUint::new(vec![1]), (1 << (0 * big_digit::BITS)));
-        check(BigUint::new(vec![N1]), (1 << (1 * big_digit::BITS)) - 1);
-        check(BigUint::new(vec![0, 1]), (1 << (1 * big_digit::BITS)));
-        check(BigUint::new(vec![N1, N1]), u64::MAX);
+        check((BigUint::one() << 64) - BigUint::one(), u64::MAX);
```
cuviper: Ditto, keep all tests.

bigint/src/lib.rs
koverstreet: On Mon, Jul 11, 2016 at 11:00:50PM -0700, Josh Stone wrote:
> >              dbits -= big_digit::BITS;
> > +            d = (c as BigDigit) >> (bits - dbits);
> 
> At first I didn't see how the `<< dbits` above could work without overflowing, but then I figured out that you're recovering those lost bits here.  This deserves some quick comments to explain.

Eh, I think the new version is just as readable as the old if you aren't just
looking at the diff.. but I can add some comments.

cuviper: That just means I should have written more comments back then, but nobody made me... :smile:

koverstreet:
I think I fixed all your comments. I renamed ilog2 to fls, and I added a correct ilog2 which is now used by the str_radix code.

bigint/src/lib.rs
```diff
@@ -1349,6 +1356,47 @@ impl Integer for BigUint {
...
+        0   => 0,
+        1   => v.data[0] as u64,
+        _   => {
+            let mut bits = v.bits();
+            let mut ret = 0u64;
+            let mut ret_bits = 0;
+
+            for d in v.data.iter().rev() {
+                let digit_bits = (bits + big_digit::BITS - 1) %
+                    big_digit::BITS + 1;
```
cuviper: Why are you adding `BITS` first?  That should be a modular no-op.  Only reason I can think is if `bits` were 0, to avoid subtraction underflow, but that should never be the case AFAICT.

bigint/src/lib.rs
cuviper: This wasn't addressed in your update.

koverstreet: Whoops, fixed.

bigint/src/lib.rs
koverstreet: bits shouldn't be 0 but IMO it's better and more readable to do it the way
that's correct in general, and post optimizer there's no difference anyways.

cuviper: Strictly speaking, it's not correct.  It just trades a should-be-impossible subtraction error for a possible-but-unlikely addition overflow.  And neither way gives a reasonable result, because `bits==0` is a corner case where `digit_bits` should be 0, otherwise `bits -= bits_want` will break.  It's not even more readable, because you had to wrap the line to fit the extra addition.

koverstreet: Hm. Actually, if adding to bits can overflow, bits itself can overflow -
it's a size_t, but size of an array in bits isn't guaranteed to fit in a
size_t.

On Jul 14, 2016 9:10 PM, "Josh Stone" <notifications@github.com> wrote:

In bigint/src/lib.rs
<https://github.com/rust-num/num/pull/204#discussion_r70922274>:

> @@ -1349,6 +1356,47 @@ impl Integer for BigUint {
>      }
>  }
>
> +fn high_bits_to_u64(v: &BigUint) -> u64 {
> +    match v.data.len() {
> +        0   => 0,
> +        1   => v.data[0] as u64,
> +        _   => {
> +            let mut bits = v.bits();
> +            let mut ret = 0u64;
> +            let mut ret_bits = 0;
> +
> +            for d in v.data.iter().rev() {
> +                let digit_bits = (bits + big_digit::BITS - 1) %
> +                    big_digit::BITS + 1;

Strictly speaking, it's not correct. It just trades a should-be-impossible
subtraction error for a possible-but-unlikely addition overflow. And
neither way gives a reasonable result, because bits==0 is a corner case
where digit_bits should be 0, otherwise bits -= bits_want will break. It's
not even more readable, because you had to wrap the line to fit the extra
addition.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rust-num/num/pull/204/files/a1772f2a0d3e8fb7384a89322c93816b017fcc57#r70922274>,
or mute the thread

cuviper: Yes.  On 32-bit platforms, a `BigUint` that grows more than 512MB is going to have a bad time anywhere that's using `::bits()`.  One that's just on the edge could create the possible-but-unlikely addition overflow here.  Meh.

koverstreet: That's kind of how I feel :)

I can go back to the subtract, I just want to get this patch in and get to
more interesting stuff...

cuviper:
@homu r+ 8e0baec

Issue #200, opened 2016-06-29
BigUint compares poorly with Python-3's int 


saurvs:
Timing trivial implementations, in Rust 1.9 and Python 3, for generating the Fibonacci series shows a stark difference in performance.

```rust
extern crate num;

use num::*;

fn fib(n: u64) -> BigUint {
    if n == 1 || n == 2 {
        return BigUint::one();
    }

    let mut a = BigUint::one();
    let mut b = BigUint::one();

    for _ in 0..n {
        b = b + &a;
        a = &b - &a;
    }

    b
}

fn main() {
    fib(100000);
}
```

```bash
$ cargo build --release
$ time ./target/debug/fib

real	0m28.996s
user	0m27.539s
sys	    0m0.097s
```

```python
def fib(n):
    if n == 1 or n == 2:
        return 1
    a = 1
    b = 1
    for _ in range(n):
        b = b + a
        a = b - a
    return b

fib(100000)
```

```bash
$ time python3 fib.py

real	0m0.466s
user	0m0.444s
sys	    0m0.012s
```

bluss:
Can you please repeat that using optimizations? Run the binary in target/release

ssrvv:
```bash
$ time ./target/release/fib

real	0m0.633s
user	0m0.612s
sys     0m0.005s
```

Okay, still slower than Python.

Timing both programs with `fib(1000000)` (extra zero) gave:

```bash
$ time ./target/release/fib

real	1m7.219s
user	1m2.668s
sys     0m0.277s
```

```bash
$ time python3 fib.py

real	0m40.180s
user	0m38.090s
sys     0m0.143s
```

cuviper:
Just to be clear, this isn't a performance win for the Python language itself.  Most of its time here is spent in `x_add` and `x_sub`, which are written in C.  The top 5 profiled for me are:

```
   51.09%  python3  libpython3.5m.so.1.0  [.] x_sub
    46.50%  python3  libpython3.5m.so.1.0  [.] x_add
     0.33%  python3  libpython3.5m.so.1.0  [.] PyEval_EvalFrameEx
     0.13%  python3  libc-2.23.so          [.] _int_malloc
     0.11%  python3  libpython3.5m.so.1.0  [.] _PyObject_Alloc.isra.0
```

The top 5 I get for your Rust version are:

```
[...]

PR #205, opened 2016-07-09
bigint: allow `Sub` to work in-place on the RHS

cuviper:
A new Fibonacci benchmark demonstrates the improvement by using both
addition and subtraction in each iteration of the loop, like #200.

Before:

    test fib2_100          ... bench:       4,558 ns/iter (+/- 3,357)
    test fib2_1000         ... bench:      62,575 ns/iter (+/- 5,200)
    test fib2_10000        ... bench:   2,898,425 ns/iter (+/- 207,973)

After:

    test fib2_100          ... bench:       1,973 ns/iter (+/- 102)
    test fib2_1000         ... bench:      41,203 ns/iter (+/- 947)
    test fib2_10000        ... bench:   2,544,272 ns/iter (+/- 45,183)

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -40,12 +40,25 @@ fn factorial(n: usize) -> BigUint {
     f
 }
 
+/// Compute Fibonacci numbers
 fn fib(n: usize) -> BigUint {
     let mut f0: BigUint = Zero::zero();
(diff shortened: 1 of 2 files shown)


hauleth:
@homu r+

PR #207, opened 2016-07-15
bigint: Break out into multiple files

koverstreet:


bigint/src/biguint_tests.rs
```diff
@@ -0,0 +1,1236 @@
...
+#[test]
+fn test_mul_divide_torture() {
+    test_mul_divide_torture_count(1000);
+}
+
+#[test]
+#[ignore]
+fn test_mul_divide_torture_long() {
+    test_mul_divide_torture_count(1000000);
+}
```
hauleth: This file should be moved to `test` directory.

koverstreet: I suppose I can do that.

cuviper:
If the tests can be compiled outside altogether, then make the `tests/` directory a sibling of `src/`.  Cargo will pick up each one automatically.  This isn't always possible though, if a test uses private items.

koverstreet:
They do use private items

hauleth:
The best would be splitting it into 2 kinds of tests:

- unit tests that checks simple methods
- integration tests that checks interface as a whole

> Wiadomość napisana przez koverstreet <notifications@github.com> w dniu 16.07.2016, o godz. 03:53:
> 
> They do use private items
>

koverstreet:
Perhaps, but let's leave that sort of thing off for future patches. I want
to keep this patch at just simple rearranging of code.

cuviper:
@homu r+ 2795223

PR #215, opened 2016-08-05
Add `is_one` method to `One` trait.

### LINKED ISSUE #214: Add method `is_one` to `One` trait
The [`Zero` trait](http://rust-num.github.io/num/num/trait.Zero.html) contains two methods: `zero` returns the number 0 for the given type and `is_zero` is provided and can be used to check if a number is 0.

The [`One` trait](http://rust-num.github.io/num/num/trait.One.html) is similar but contains only the method `one` to create the number 1. Please also add a method to check if a number is one.

pyfisch:
Implement `is_one` for all number types in crate.

Closes #214

BREAKING CHANGE: Implementors of `One` need to provide
a `is_one` method.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -613,7 +613,7 @@ impl<T: Clone + Num> One for Complex<T> {
     
     #[inline]
     fn is_one(&self) -> bool {
-        self.re == One::one() && self.im == Zero::zero()
+        self.re.is_one() && self.im.is_zero()
     }
 }
 
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -610,6 +610,11 @@ impl<T: Clone + Num> One for Complex<T> {
     fn one() -> Complex<T> {
         Complex::new(One::one(), Zero::zero())
     }
+    
+    #[inline]
+    fn is_one(&self) -> bool {
+        self.re == One::one() && self.im == Zero::zero()
+    }
 }
 
 /* string conversions */
--- a/rational/src/lib.rs
+++ b/rational/src/lib.rs
@@ -476,6 +476,11 @@ impl<T: Clone + Integer> One for Ratio<T> {
     fn one() -> Ratio<T> {
         Ratio::new_raw(One::one(), One::one())
     }
+    
+    #[inline]
+    fn is_one(&self) -> bool {
+        self == &Ratio::one()
+    }
 }
 
 impl<T: Clone + Integer> Num for Ratio<T> {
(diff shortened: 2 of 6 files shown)


cuviper:
Hmm, right. I thought we could have a default implementation, but not without `eq`. It's one more thing to add to the growing breaking-change list.

complex/src/lib.rs
```diff
@@ -610,6 +610,11 @@ impl<T: Clone + Num> One for Complex<T> {
     fn one() -> Complex<T> {
         Complex::new(One::one(), Zero::zero())
     }
+    
+    #[inline]
+    fn is_one(&self) -> bool {
+        self.re == One::one() && self.im == Zero::zero()
```
cuviper: Can't this use `is_one` and `is_zero`?

rational/src/lib.rs
```diff
@@ -476,6 +476,11 @@ impl<T: Clone + Integer> One for Ratio<T> {
     fn one() -> Ratio<T> {
         Ratio::new_raw(One::one(), One::one())
     }
+    
+    #[inline]
+    fn is_one(&self) -> bool {
+        self == &Ratio::one()
```
cuviper: Perhaps `numer == denom`, assuming a non-zero denominator.

complex/src/lib.rs
pyfisch: It can.

rational/src/lib.rs
pyfisch: done.

hauleth:
I think that for such change we should wait for rust-lang/rust#31844

cuviper:
How would you envision specialization helping here?

hauleth:
Using `default` keyword: https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md#the-default-keyword

hauleth:
But it is already possible to make it non-breaking change https://is.gd/zKpGwk

cuviper:
Ah, I missed the fact that `default impl` can be partial.  In that case, yes we could provide `is_one` for `T: Eq`, but it would still be a breaking change for any `T` lacking that.

Your non-breaking example makes `Foo::b()` completely depend on `Eq` -- generic callers could only use this method if they require `T:Foo+Eq` themselves.  Some `Eq` types could choose to provide a better implementation, but non-`Eq` types can never implement this.

cuviper:
I picked this into the `next` branch in #336, thanks!

PR #219, opened 2016-08-11
Inline small functions, especially wrappers

### LINKED ISSUE #218: Inline small wrapper functions
Such as the `PrimInt` implementations for built-in integers.

See https://github.com/briansmith/ring/pull/199#issuecomment-238817368

cuviper:
We already had `#[inline]` throughout a lot of the code, but notably some
functions which simply wrap inherent methods were not inlined.  That means
external references will get a full function call, when they could have been
optimized to as little as one opcode.

This PR inlines all functions that look tiny enough for this to matter.

Fixes #218.


hauleth:
@homu r+

Issue #194, opened 2016-06-08
Integrate common floating point constants


b52:
In most of my implementations involving generic floating point values I have the need to work with pi and/or e. Thus it would be really helpful to provide static functions to retrieve those values for the concrete type.

cuviper:
In general, `let e = Float::one().exp();` and  `let pi = (-Float::one()).acos();` should work.  We could provide these as default implementations and override the specific constants for `f32` and `f64`.

cuviper:
Associated constants would be nice too, rust-lang/rust#29646, but we couldn't provide defaults there.

IvanUkhov:
I agree that pretty much every mathematical constant can be computed using the methods that are already there. However, `(-Float::one()).acos()` and alike obscure the actual algorithm. Do we really want to see people in the Rust community writing this kind of generic code? Besides, who is willing to guarantee that `(-Float::one()).acos()` is indeed a zero-cost abstraction?

cuviper:
No, I don't expect people should do that -- it's just a way to write a generic default so it's not a breaking change.  We can provide the proper constants in the implementation for the concrete float types.

Care to write up a PR?

IvanUkhov:
@cuviper, first of all, I apologize if it sounded rude; no offense was intended. I’m an active user of `num` and really appreciate your work.

Yes, absolutely. If there are no objections, I’ll be happy to help.

cuviper:
No problem, sorry if I sounded too defensive. :)

cuviper:
Fixed by #220 -- thanks @IvanUkhov!

PR #220, opened 2016-08-11
Add a trait for floating-point constants

IvanUkhov:
The pull request is to address issue #194. In order to keep the library organized, I’ve introduced a new trait for the new functionality. The trait is supposed to closely follows the [`consts`](https://doc.rust-lang.org/std/f64/consts/index.html) module from the standard library. There are at least the following three open questions:

1. What should the name of the trait be? Currently, it’s `Constant`; an alternative is `Consts`.
2. What should the names of the getters be? Currently, they are lower-case versions of the constants defined in the `consts` module. Note that `Float` provides `log2` and `log10`, whereas `LOG_2` and `LOG_10` get translated into `log_2` and `log_10`, respectively.
3. Should `Float` require the new trait? Or should it be left up to the user?

Please let me know what you think. Thank you!

Regards,
Ivan

### DIFF
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -1233,6 +1233,61 @@ fn integer_decode_f64(f: f64) -> (u64, i16, i8) {
 float_impl!(f32 integer_decode_f32);
 float_impl!(f64 integer_decode_f64);
 
+macro_rules! float_const_impl {
+    ($(#[$doc:meta] $constant:ident,)+) => (
+        #[allow(non_snake_case)]
+        pub trait FloatConst {
+            $(#[$doc] fn $constant() -> Self;)+
+        }
+        float_const_impl! { @float f32, $($constant,)+ }
+        float_const_impl! { @float f64, $($constant,)+ }
+    );
+    (@float $T:ident, $($constant:ident,)+) => (
+        impl FloatConst for $T {
+            $(
+                #[inline]
+                fn $constant() -> Self {
+                    ::std::$T::consts::$constant
+                }
+            )+
+        }
+    );
+}
+
+float_const_impl! {
+    #[doc = "Return Euler’s number."]
+    E,
+    #[doc = "Return `1.0 / π`."]
+    FRAC_1_PI,
+    #[doc = "Return `1.0 / sqrt(2.0)`."]
+    FRAC_1_SQRT_2,
+    #[doc = "Return `2.0 / π`."]
+    FRAC_2_PI,
+    #[doc = "Return `2.0 / sqrt(π)`."]
+    FRAC_2_SQRT_PI,
+    #[doc = "Return `π / 2.0`."]
+    FRAC_PI_2,
+    #[doc = "Return `π / 3.0`."]
+    FRAC_PI_3,
+    #[doc = "Return `π / 4.0`."]
+    FRAC_PI_4,
+    #[doc = "Return `π / 6.0`."]
+    FRAC_PI_6,
+    #[doc = "Return `π / 8.0`."]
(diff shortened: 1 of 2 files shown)


cuviper:
Thanks for this!

> What should the name of the trait be? Currently, it’s Constant; an alternative is Consts.

I don't like either name if we export the at the top level of `traits` -- it's *too* general.  Think also about how people should be expected to invoke this.  Something like `num_traits::float::Constant::e()`?  Or perhaps `use num_traits::FloatConsts;` then `FloatConsts::e()`?

> What should the names of the getters be? Currently, they are lower-case versions of the constants defined in the consts module. Note that Float provides log2 and log10, whereas LOG_2 and LOG_10 get translated into log_2 and log_10, respectively.

I guess the ones you mean are `ln_2` and `ln_10`, so the names aren't *quite* so close.  But I think we could justify all-caps function names here to mimic true constants.  (Just add an `#[allow]` for whatever lint complains.)  I don't feel strongly either way though.

> Should Float require the new trait? Or should it be left up to the user?

The nice thing about having them totally separate is that this isn't a breaking change to add it.  If `Float` requires it, then any (hypothetical) outside implementation would be broken by not having it.

IvanUkhov:
**Question 1**. I agree that the name _Constant_ is too general when considered in the context of the root module. I re-exported the trait in order to follow the conventions that I saw in the code: you keep all the modules public but at the same time re-export all their contents. If we don’t re-export `Constant`, we won’t have this problem; it’ll only be accessible through `float`.

Let’s see how different versions might look:

```rust
use num_traits::float::{Float, Const};
fn foo1<T>(bar: T) -> T where T: Float + Const {
    bar + T::pi() // or Const::pi()
}

use num_traits::float::{Float, Constant};
fn foo2<T>(bar: T) -> T where T: Float + Constant {
    bar + T::pi() // or Constant::pi()
}

use num_traits::{Float, FloatConst};
fn foo3<T>(bar: T) -> T where T: Float + FloatConst {
    bar + T::pi() // or FloatConst::pi()
}

use num_traits::{Float, FloatConsts};
fn foo4<T>(bar: T) -> T where T: Float + FloatConsts {
    bar + T::pi() // or FloatConsts::pi()
}

use num_traits::{Float, FloatConstant};
fn foo5<T>(bar: T) -> T where T: Float + FloatConstant {
    bar + T::pi() // or FloatConstant::pi()
}

use num_traits::{Float, FloatConstants};
fn foo6<T>(bar: T) -> T where T: Float + FloatConstants {
    bar + T::pi() // or FloatConstants::pi()
}
```

To me personally, singular forms make more sense. Regarding with or without `float::`, probably it’s better to make the new trait directly accessible, just like all other items in the crate; hence, `Float` should be in the name of the new trait. In this case, `FloatConstant` is getting a bit too long, so I’d personally choose `FloatConst` (`foo3`).

**Question 2**. Yes, you’re right. Sorry for the confusion. I very much like your idea about all caps.

**Question 3**. Yes, it’s better to keep `Float` independent of the constants.

You’re in charge here, and all the choices are entirely up to you. Please tell me what changes should be made to the pull request. Thank you.

cuviper:
Somehow I forgot that you can use `T::method()` for any trait that `T` implements, even though I write like that all the time.  It certainly makes this more ergonomic.

Sounds like a reasonable consensus: `FloatConst` as an independent trait with uppercase names.

IvanUkhov:
I’ve updated the pull request. Thank you.

cuviper:
Looks good, thanks!

@homu r+ 01aad70

PR #222, opened 2016-08-15
Use serde 0.8

SuperFluffy:
I updated `bigint`, `complex`, and `rational` to use `serde 0.8`, and also fixed deserialization and the `serde` feature as such in the `rational` crate (didn't add any tests, but it compiles now).

Similar to https://github.com/rust-num/num/pull/196 for `num/complex`, “`use serde;`” needed to be removed in `num/rational`.

### DIFF
--- a/.travis/test_nightly.sh
+++ b/.travis/test_nightly.sh
@@ -5,3 +5,6 @@ set -ex
 cargo bench --verbose
 
 cargo test --verbose --manifest-path=macros/Cargo.toml
+
+# Build test for the serde feature
+cargo build --verbose --features "serde"
--- a/.travis/test_nightly.sh
+++ b/.travis/test_nightly.sh
@@ -8,3 +8,7 @@ cargo test --verbose --manifest-path=macros/Cargo.toml
 
 # Build test for the serde feature
 cargo build --verbose --features "serde"
+
+# Downgrade serde and build test the 0.7.0 channel as well
+cargo update -p serde --precise 0.7.0
+cargo build --verbose --features "serde"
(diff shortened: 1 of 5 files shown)


cuviper:
The `rational` fixes are welcome right away.  The `serde` upgrade changes our public interface though, so that's Yet Another Breaking Change.  We should do that eventually, but let's separate this.

cuviper:
Hmm, since it didn't require any code change, maybe our `serde` dependency could use an explicit range.  The caret requirement uses strict semver, but we might allow something like `>= 0.7 < 0.9`.  Then we should be compatible with any user's `0.7.*` or `0.8.*`.  What do you think?

cuviper:
We really should have at least a build test too.  Can you add a line to `.travis/test_nightly.sh`?
(Only nightly just because I know serde won't work with our rust-1.0.0 build.)

SuperFluffy:
Alright, the PR was split into two, see https://github.com/rust-num/num/pull/223. I am still basing the `serde` update off of the other one, since building the entire `num` crate with the `serde` feature is broken otherwise.

cuviper:
I confirmed that the dual-serde works locally, but could you add another test for this?  The existing test you added will grab the latest possible, so that will be some 0.8.*.  After that, you can force a downgrade like `cargo update -p serde --precise 0.7.0` and build again.

SuperFluffy:
All done, thanks for the extra requests. I finally had a reason to mess around with travis a bit. :-)

cuviper:
Great, thanks!

@homu r+ 4f6f7b3

SuperFluffy:
What is the process of getting a new version of this (`0.1.35`?) to `crates.io`?

cuviper:
I just need to go through and publish each update crate.  I'll try to do that tonight.

cuviper:
OK, 0.1.35 is out, enjoy!

PR #223, opened 2016-08-17
Fix import serde and rational deserialization

SuperFluffy:
Similar to #196 for num/complex, “use serde;” needed to be removed in num/rational.

Also, deserialization of `num/rational` needed to be fixed by adding type annotations.

This is in response to https://github.com/rust-num/num/pull/222#issuecomment-239957062 of issue https://github.com/rust-num/num/pull/222.

Also added a travis line, in response to https://github.com/rust-num/num/pull/222#issuecomment-239957062. Hope it works.

### DIFF
--- a/.travis/test_nightly.sh
+++ b/.travis/test_nightly.sh
@@ -5,3 +5,6 @@ set -ex
 cargo bench --verbose
 
 cargo test --verbose --manifest-path=macros/Cargo.toml
+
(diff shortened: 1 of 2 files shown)


cuviper:
Thanks!

@homu r+ eb486e8dc149cbd49c7c111860079430f8906358

PR #224, opened 2016-09-16
Add BigDecimal Crate

akubera:
Rebase of pull request #193. Addresses issue #8.

Implementation of BigDecimal, which comprises a BigInt and a u64 scale factor, allowing for storage of 2^64 digits.

Further implementation is required to set the default maximum precision (so numbers like 1/3 are not represented by 2^64 digits). The current default is 100.

Add, Sub, Mul implemented by @iterion, I reimplemented Div, Eq, FromStr.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ bigint = ["num-bigint"]
 bigdecimal = ["num-bigdecimal"]
 complex = ["num-complex"]
 rational = ["num-rational"]
-default = ["bigint", "complex", "rational", "rustc-serialize"]
+default = ["bigdecimal", "bigint", "complex", "rational", "rustc-serialize"]
 
 serde = [
   "num-bigint/serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,11 @@ optional = true
 path = "bigint"
 version = "0.1.35"
 
+[dependencies.num-bigdecimal]
+optional = true
+path = "bigdecimal"
+version = "0.1.32"
+
 [dependencies.num-complex]
 optional = true
 path = "complex"
@@ -53,6 +58,7 @@ version = "0.3.8"
 
 [features]
 bigint = ["num-bigint"]
+bigdecimal = ["num-bigdecimal"]
 complex = ["num-complex"]
 rational = ["num-rational"]
 default = ["bigint", "complex", "rational", "rustc-serialize"]
--- a/bigdecimal/src/lib.rs
+++ b/bigdecimal/src/lib.rs
@@ -43,7 +43,7 @@ use std::str::{self, FromStr};
 use std::fmt;
 use std::cmp::Ordering::Equal;
 use std::cmp::max;
-use bigint::{BigInt, ParseBigIntError};
+use bigint::{BigInt, BigUint, ParseBigIntError, Sign};
 use traits::{Num, Zero, One};
 // use num_traits::identities::Zero;
 
@@ -157,10 +157,10 @@ impl BigDecimal {
 
         if new_scale > self.scale {
             let raise = new_scale - self.scale;
-            println!("raise: {}", raise);
+            // println!("raise: {}", raise);
             let exp = (10 as i64).pow(raise as u32);
             let new = BigDecimal::new(self.int_val.clone() * BigInt::from(exp), new_scale);
-            println!("new: {}", new);
+            // println!("new: {}", new);
             return new;
         }
 
@@ -333,15 +333,89 @@ impl Num for BigDecimal {
     /// Creates and initializes a BigDecimal.
     #[inline]
     fn from_str_radix(s: &str, radix: u32) -> Result<BigDecimal, ParseBigDecimalError> {
-        let scale = match s.find('.') {
-            Some(i) => (s.len() as i64) - (i as i64) - 1,
+        assert!(2 <= radix && radix <= 36, "The radix must be within 2...36");
+
+        // interpret as characters
+        let mut chars = s.chars().peekable();
+
+        // get the value of the first character
+        let firstchar = match chars.peek() {
+            Some(&c) => c,
+
+            // empty string returns zero
+            None => return Ok(BigDecimal::zero()),
+        };
+
+        // is this a negative number
+        let sign = if firstchar == '-' {
+            Sign::Minus
+        } else {
+            Sign::Plus
+        };
+
+        // number of leading characters to ignore
+        let offset = match firstchar {
+            '+' | '-' => 1,
+            '0'...'9' | '.' => 0,
+            _ => panic!("Unexpected beginning character"),
+        };
+
+        // storage buffer
+        let mut buff = Vec::with_capacity(s.len());
+
+        // this optional decimal point
+        let mut decimal_found = None;
+
+        // keep track of double underscores
+        let mut prev_underscore = false;
+        for (i, c) in chars.skip(offset).enumerate() {
+
+            if c == '_' && prev_underscore {
+                panic!("Malformed string (multiple _ characters)");
+            } else if c == '_' {
+                prev_underscore = true;
+                continue;
+            } else {
+                prev_underscore = false;
+            }
+
+            //
+            if c == '.' {
+                if decimal_found == None {
+                    decimal_found = Some(i);
+                    continue;
+                } else {
+                    panic!("Multiple decimal points");
+                }
+            }
+
+            // foo
+            let d = match c {
+                '0'...'9' | 'a'...'f' | 'A'...'F' => c as u8,
+                // Some(digit) => buff.push(digit),
+                _ => panic!("Unexpected character {:?}", c),
+            };
+
+            // println!("{}", c.to_digit());
+            buff.push(d);
+        }
+
+        let scale = match decimal_found {
+            Some(val) => (s.len() as u64) - (val as u64) - 1,
             None => 0,
         };
 
-        match BigInt::from_str_radix(s, radix) {
-            Ok(bi) => Ok(BigDecimal::new(bi, scale as i64)),
-            Err(_) => Err(ParseBigDecimalError::Other), // ParseDecimal(error)),
-        }
+        let big_uint = match BigUint::parse_bytes(&buff, radix) {
+            Some(x) => x,
+            None => BigUint::zero(),
+        };
+
+
+        let big_int = BigInt::from_biguint(sign, big_uint);
+        return Ok(BigDecimal {
+            int_val: big_int,
+            scale: scale as i64,
+        });
     }
 }
 
@@ -349,7 +423,9 @@ impl Num for BigDecimal {
 mod bigdecimal_tests {
     use super::BigDecimal;
     use bigint::BigInt;
-    use traits::Num;
+    use traits::{Num, ToPrimitive};
+
+    use std::str::FromStr;
 
     /// Assert that an op works for all val/ref combinations
     macro_rules! assert_op {
@@ -405,18 +481,18 @@ mod bigdecimal_tests {
 
     #[test]
     fn test_from_str() {
-        let a = BigDecimal::from_str_radix("1", 10).unwrap();
+        let vals = vec![
+            ("1331.107", 1331107, 3),
+            ("1.0", 10, 1),
+            ("0.00123", 123, 5),
+            ("-123", -123, 0),
+            ("-1230", -1230, 0),
+        ];
 
-        // for &(x, y) in [("100.4", "1.0000"), ("4.5", "4.50")].iter() {
-        //     println!("X: {}", x);
-        //     let a = BigDecimal::from_str_radix(x, 10);
-        //     match a {
-        //         Ok(_) => println!("OK"),
-        //         Err(_) => println!("Err"),
-        //     }
-        //     // println!(">> {}", a);
-        //     a.unwrap();
-        //     // let b = BigDecimal::from_str_radix(y, 10).unwrap();
-        // }
+        for &(source, val, scale) in vals.iter() {
+            let x = BigDecimal::from_str(source).unwrap();
+            assert_eq!(x.int_val.to_i32().unwrap(), val);
+            assert_eq!(x.scale, scale);
+        }
     }
 }
--- a/bigdecimal/src/lib.rs
+++ b/bigdecimal/src/lib.rs
@@ -171,6 +171,7 @@ impl BigDecimal {
 #[derive(Debug, PartialEq)]
 pub enum ParseBigDecimalError {
     ParseDecimal(ParseFloatError),
+    Empty,
     Other,
 }
 
@@ -178,6 +179,7 @@ impl fmt::Display for ParseBigDecimalError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             &ParseBigDecimalError::ParseDecimal(ref e) => e.fmt(f),
+            &ParseBigDecimalError::Empty => "Failed to parse empty string".fmt(f),
             &ParseBigDecimalError::Other => "failed to parse provided string".fmt(f),
         }
     }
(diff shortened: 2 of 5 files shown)


cuviper:
I still need to sit down and do a full review of this, but since you're apparently tweaking the parsing right now -- it looks over-complicated to me.  I would actually expect this to be a rather "dumb" wrapper -- just finding the decimal point if it exists and letting `BigInt` do the real work (actually `BigUint` working under that), including your enhancement for dealing with underscores.

cuviper:
I know you *are* leaning on `BigInt` and `BigUint` to some extent, but I think most of the char-parsing is out of place.  I think it should be possible to only worry about '.' and the 'e' exponent here, and leave *everything* else to the lower layer.

akubera:
Ok, I believe a well placed `partition` method can simplify things. I'll see what I can do.

akubera:
Errr, `partition` isn't what I thought it was (python's `str.partition`). It's now a few `split_at` calls and special case handling.

Question: Should I support underscores?

akubera:
Ok, I re-read and see now that underscore support should be removed. That's been done, and the from_str_radix function I think is ready for review. There are lots of comments, I hope that's ok.

akubera:
In terms of the decimal point, I find it and (if found) push the characters before and after it into a String, and pass that to the BigInt. I think that requires a copy. I don't know if there's a better way around that (aside from a BigInt constructor from Iterator<char> and chain the two slices).

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+        // split slice into base and exponent parts
+        let (base_part, exponent_value) = match s.find(exp_separator) {
+            // exponent defaults to 0 if (e|E) not found
+            None => (s, 0),
+
+            // split and parse exponent field
+            Some(loc) => {
+                let (base, exp) = s.split_at(loc);
+                // slice after 1 to skip the 'e' char
+                (base, try!(i64::from_str(&exp[1..])))
```
cuviper: We currently still support Rust 1.0, but `split_at` wasn't added/stabilized until 1.4.

But just `s[..loc]` and `s[loc+1..]` should be fine.  (It's mutable slicing where you get into trouble.)

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+        f.pad_integral(true, "", &self.int_val.to_str_radix(10))
+    }
+}
+
+impl Num for BigDecimal {
+    type FromStrRadixErr = ParseBigDecimalError;
+
+    /// Creates and initializes a BigDecimal.
+    #[inline]
+    fn from_str_radix(s: &str, radix: u32) -> Result<BigDecimal, ParseBigDecimalError> {
```
cuviper: It's a little strange to support a radix on something called "decimal", but hey, that's what `Num` wants... :)

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+                let mut digits = String::from_str(lead).unwrap();
+
+                // copy all trailing characters after '.' into the digits string
+                digits.extend(trail.chars().skip(1));
+
+                (digits, trail.len() as i64 - 1)
+            }
+        };
+
+        let scale = decimal_offset - exponent_value;
```
cuviper: Hmm, actually on the radix point, this is wrong for radixes other than 10, and others can't be perfectly converted.
(e.g. `"0.1"` radix 3 is `1/3`)

cuviper:
On the underscore support, I'm OK with having that -- it just belongs in `BigUint`'s parser.  That can be a PR independent of this one, if you'd like a smaller task. :)

bigdecimal/src/lib.rs
bluss: `(&s[..loc], &s[loc..])`, the beauty of half open ranges.

libs people don't think bumping the rust version is breaking https://github.com/rust-lang/rfcs/pull/1619#issuecomment-240242943  Every such thing is a judgment call though, of course.

cuviper: @bluss I think they're wrong about that, and the comment you linked acknowledges that this idea was the most contentious part of that RFC.  But this is not the place to argue that.

bluss: fair enough

akubera:
Thanks for the review. Most things not `from_str_radix` and `eq` were written by @iterion, so I don't know if there was, for example, an intention to use `rand` or not.

Is adding "bigdecimal" to the Makefile all that's needed for travis testing to work?

I'll revisit the project this weekend.

hauleth:
@akubera yep, that should be everything what you need.

iterion:
@akubera no idea why I added rand, my guess was it was just a copy paste from somewhere else that got left in.

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+forward_all_binop_to_ref_ref!(impl Add for BigDecimal, add);
+
+impl<'a, 'b> Add<&'b BigDecimal> for &'a BigDecimal {
+    type Output = BigDecimal;
+
+    #[inline]
+    fn add(self, other: &BigDecimal) -> BigDecimal {
+        let scale = max(self.scale, other.scale);
+        let left = self.clone().set_scale(scale);
+        let right = other.clone().set_scale(scale);
```
akubera: The math equations would benefit from a single clone (or rather "rescale") and then an inplace operator on the new object. It looks like `std::ops::AddAssign` was marked stable in 1.8; which means I cannot use that, right? Is there a way to write two implementations based on rustc version?

cuviper:
I'm not sure what tangible benefit you'd get from AddAssign. When you have
a local object, you can just move it by value into Add and accept what
comes out.  That is, "x += y" and "x = x + y" are not much different.

bigdecimal/src/lib.rs
akubera: Do you think I should just Err if radix != 10?

cuviper: Sure, I think that's reasonable.

Cargo.toml
```diff
@@ -53,6 +58,7 @@ version = "0.3.8"
 
 [features]
 bigint = ["num-bigint"]
+bigdecimal = ["num-bigdecimal"]
 complex = ["num-complex"]
 rational = ["num-rational"]
 default = ["bigint", "complex", "rational", "rustc-serialize"]
```
akubera: ✅

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+        }
+
+        if self.int_val.is_zero() {
+            return BigDecimal::new(BigInt::zero(), new_scale);
+        }
+
+        if new_scale > self.scale {
+            let raise = new_scale - self.scale;
+            // println!("raise: {}", raise);
+            let exp = (10 as i64).pow(raise as u32);
```
akubera: This has been fixed by looping and multiplying BigInts until correct number is reached. I don't know if there's a better way of doing this. The only other way I could think of is formatting the string `BigDecimal::from_str(&format!("1e{}", raise));` which does pass my tests and probably has less room for error, but for some reason, it seems wrong.

I think it'd be nice to have a BigInt (or BigUint) `::as_order_of_magnitude(u64) -> Self` method that would handle this in the best way for us. Does something like this exist?

thibaultdelor:
What is the status on that? Rust is definitely lacking an implementation of BigDecimal.
I would be keen to work on it, I am just wondering what would be required.

akubera:
It works insofar the [tests](https://github.com/akubera/num/blob/topic/big-decimal/bigdecimal/src/lib.rs#L473) for construction and arithmetic pass, but the implementation is not as efficient as it could/should be. As it stands, it uses a BigInt to store the numerical data, so we lose access to the BigDigits which could be used for optimization for, say, testing equality.

The other thing to worry about is how to manage precision. Right now I limit decimal places in the division operator with an [arbitrary number](https://github.com/akubera/num/blob/topic/big-decimal/bigdecimal/src/lib.rs#L373). The python decimal module (I almost typed 'crate', haha) uses thread-scoped [context objects](https://docs.python.org/3/library/decimal.html#context-objects) to manage such information; that's probably not the way to go for Rust. I haven't worked with/looked at any other such library, someone should throw ideas around.

I don't know the industry standards for such a library: is a goal keeping track of significant figures? should (1.00 == 1.00000)? should there be a function that returns false in that case? Should it coupled to the num-rational crate? Should every object have their own "context" properties? What rules govern objects with different such properties?

I'd need input from experts before moving forward, and I've been busy and haven't had time to work on this for couple months. Any input would be appreciated.

rpedela:
> I don't know the industry standards for such a library: is a goal keeping track of significant figures? should (1.00 == 1.00000)? should there be a function that returns false in that case? Should it coupled to the num-rational crate? Should every object have their own "context" properties? What rules govern objects with different such properties?

I think it may be worth looking at the source for Postgres's "numeric" type. The PG community cares a lot about correctness and I am willing to bet they have tackled many of your non-Rust-specific questions.

Github mirror: https://github.com/postgres/postgres
hacker mailing list: https://www.postgresql.org/list/pgsql-hackers/

rubdos:
> ☔️ The latest upstream changes (presumably #263) made this pull request unmergeable. Please resolve the merge conflicts.

[I rebase the thing on master](https://github.com/rubdos/num/tree/big-decimal)

rubdos:
> I think it may be worth looking at the source for Postgres's "numeric" type. The PG community cares a lot about correctness and I am willing to bet they have tackled many of your non-Rust-specific questions.

IRC:

> 2017-04-04 11:25:22	rubdos	Someone knows out of the top of their head whether postgresql considers NUMERIC 1.00000 == 1.0 ?
> 2017-04-04 11:25:37	johto         yes

iterion:
@rubdos yeah, a new PR is how we originally handled it when I ran out of time to work on this. So, 👍 from me. Would still love to see this merged as well as BigDecimal support for diesel, which is why I originally started this!

akubera:
Yes, that is fine with me.

cuviper:
I think what might be best for this is to develop `BigDecimal` as a standalone project.

There's clearly a lot of interest in having this type, and I feel bad about dragging my feet.  But at the same time, there are still design questions that *I'm* not terribly comfortable with, and don't have great ideas how to make it better.  Yet we rust-num maintainers are the ones stuck with the maintenance bill if we merge this.  Plus, we treat `num` stability very conservatively, which doesn't mesh well with a crate that may still need nascent development iteration.

So, how would you folks feel about starting a separate project for `BigDecimal`?  I think the only thing you'd lose is getting a re-export from the base `num` crate, but IMO that's not such a big deal.  Is there any reason you think this *needs* to be part of `rust-num/num`?

akubera:
Other than relieving people of the paradox of choice, I don't think it would *need* to be in num.

I think it'd be fine to have a few competing implementations, and perhaps one day there would be one that is unarguably better than the others, and be blessed by the num developers and merged into the project. I certainly like having a "one-stop-shop" crate handling multiple similar needs, and I would trust that the maintainers of such crates have worked on ensuring quality and interoperability.

I can think of two options:

1) Let anyone interested (@rubdos) create their own repository and work on their own. Maybe `rust-num/num` keeps an issue open keeping track of the status of participating projects. 
2) A rust-num owner could create a `rust-num/bigdecimal` or `rust-num/bigdecimal-nursery` repository, with different implementations managed on different branches. Tradeoffs and designs can be discussed & implemented in this repo, and collectively people learn what works and what doesn't.

2 is obviously more difficult, but would provide people who know what they're doing an easy-to-find place to comment, and would reduce reinvention of the wheel for anyone who wants to start something new.
Also, providing a standard testing+benchmark suite would be nice.

iterion:
Overall, I'm cool with the idea of separating this out into its own package. Ultimately, if the implementation stabilizes then rust-num/num would be a great place for it. But, we're not there yet and it also assumes that a single standard can be agreed upon. That said, it seems other languages make that choice for you: [ruby](https://ruby-doc.org/stdlib-1.9.3/libdoc/bigdecimal/rdoc/BigDecimal.html), [python](https://docs.python.org/2/library/decimal.html), etc.

I like some form of @akubera's option 2, as it doesn't necessarily preclude competing standards while still making it easy for newcomers to find some sort of community agreed upon decimal implementation. I'm happy to volunteer to help maintain any project that is created as well.

rubdos:
It looks like `Add` is bugged; I'm getting invoices of a few trillion euro's :D

rubdos:
> Is there any reason you think this needs to be part of rust-num/num?

Yes. Obviously because rust-num is **the** place to be for everything number related, and I think avoiding separation would enhance potential cohesion of the Rust ecosystem.

I'm totally okay with splitting off; the only remaining question will be whether Diesel would still merge something that pulls from `rubdos/big-decimal-rs` instead of `rust-num`. Btw, the Diesel implementation is linked in somewhere here, and I must say that it's pretty neat.

rubdos:
You know what; I'll make the merge request for a split-off crate. You guys leave it open, review from now and then, and in a few months, when things stabilize, we merge again. In the meantime, I'll maintain a separate crate for this alone. Sounds like an idea?

cuviper:
I think you're putting too much emphasis on rust-num's "blessing" -- we're just a couple of people who volunteered to maintain `num` when it was booted from the standard library.  Beyond that, there's no connection to rust-lang proper as far as having some sort of official sanctioning power.  But maybe I underestimate the network effects, who knows.

The separate project doesn't need to be treated as a competition, unless you folks disagree about something between yourselves.  Just have someone open a personal repo and invite the others to collaborate.  Later down the road, we can evaluate whether it makes sense to merge into `rust-num/num`, move it into `rust-num/big-decimal`, or just continue living on its own.

@rubdos if you really want to keep an open merge request here, I guess that's ok too.

sgrif:
Someone please release a crate somewhere so I have a type to target for the PG decimal type. I'm sick of telling people that I'm not going to implement `FromSql<Decimal> for f64`

akubera:
I've made a bigdecimal repository maintaining the history of the subdirectory here: [akubera/bigdecimal-rs](github.com/akubera/bigdecimal-rs), I assume @iterion is ok with that. I've ignored the 1.0.0 compatible commits, so I only know it's compatible with stable (1.16.0) and beyond. @rubdos, you're free to clone (i.e. not fork but hard clone) and use however you and the Desiel project wishes (I started cherry-picking before I realized you said you'd fork it).

rubdos:
> I've made a bigdecimal repository maintaining the history of the subdirectory here: akubera/bigdecimal-rs, I assume @iterion is ok with that. I've ignored the 1.0.0 compatible commits, so I only know it's compatible with stable (1.16.0) and beyond. @rubdos, you're free to clone (i.e. not github fork but hard clone and push to your own repo) and use however you and the Desiel project wishes (I started cherry-picking before I realized you said you'd fork it).

I'd rather work together then. Want to meet up on an IRC channel, or shall I open some issues on your repo?

cuviper:
I found a few existing crates too:

- [decimal](https://crates.io/crates/decimal): fixed 128-bit representation
- [rust_decimal](https://crates.io/crates/rust_decimal): also fixed 128-bit
- [decimate](https://crates.io/crates/decimate): generic significand, could be `BigInt`

I'd take a close look at `decimate` then...

rubdos:
It's probably worth noticing that code based on @iterion's work has been pushed to the `bigdecimal` crate, and is further developed at [@akubera 's](/akubera/bigdecimal-rs) repo

cuviper:
Cool - happy hacking everyone!

bigdecimal/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+
+    #[inline]
+    fn from_str(s: &str) -> Result<BigDecimal, ParseBigDecimalError> {
+        BigDecimal::from_str_radix(s, 10)
+    }
+}
+
+impl PartialEq for BigDecimal {
+    #[inline]
+    fn eq(&self, rhs: &BigDecimal) -> bool {
```
legolord208: Java converts both arguments to BigIntegers and compares those.... Yeah that's even worse, isn't it?

cuviper:
I'm trying to clean up old PRs.  Since you all have moved to a separate crate, I think that it's not helpful to leave this open.  Hope it's going well!

PR #228, opened 2016-09-18
Fix `num-macros` `FromPrimitive` implementation

### LINKED ISSUE #227: Make `num-macros` use syn crate for compatibility with Macros 1.1
When [Macros 1.1](https://github.com/rust-lang/rfcs/blob/master/text/1681-macros-1.1.md) lands in stable we would be ready to provide `num-macros` there also. This will mean that we need somehow parse Rust code and `syntex` crate is quite slow. However there is [`syn`](https://github.com/dtolnay/syn) crate that use `nom` and is quite fast.

hauleth:
Current solution follow syntax proposed in rust-lang/rfcs#1681.

Tracking issue rust-lang/rust#35900

Close #227

Code broken after change:

- [ ] https://github.com/kdar/gameboy-rs
- [ ] https://github.com/lemonrock/mbedtls
- [ ] https://github.com/timorantalaiho/kelaa
- [ ] https://github.com/dylanede/cef-rs
- [ ] https://github.com/klutzy/aheui-llvm
- [ ] https://github.com/pcwalton/rust-media
- [ ] https://github.com/PistonDevelopers/hematite_server


hauleth:
Unfortunately due to bug dtolnay/syn#12 this syntax will currently fail:

```rust
#[derive(FromPrimitive)]
enum Foo {
    A = 1,
    B
}
```

hauleth:
cc @cuviper

cuviper:
Does the existing code work with that problem case? (Apart from the current nightly break.). If so, I think we should try to fix that break and add a test like that.  Then when syn releases a fix for that bug, we can land this PR.

hauleth:
Existing code doesn't work at all as there was breaking change in compiler plugins (again) and that made `num-macros` uncompilable. Current version compile and works except that one case.

cuviper:
That's what I meant by the current nightly break. Can we fix that without this whole rewrite, then switch when syn is ready?

hauleth:
I haven't written compiler plugin even once so I am not sure if I can fix this. About this issue I can use syntex crate but this will significantly increase compile times for `num-macros`.

bluss:
Syntax extensions are annoying for now that they break so often, but the fixes are usually very simple to make.

hauleth:
@bluss probably it is just fixing the name, but since rust-lang/rfcs#1681 was accepted I think we should jump into that train and use Macros 1.1 syntax. As you can see in the PR it makes whole code much simpler. The only problem is that `syn` crate need a little fix (which I cannot make because I am not the `nom` wizard and I always forget how to write parsers in it. However it should be also quite simple task.

hauleth:
We can always drop `syn` and use `syntex` instead, but AFAIK `syntex` compile times aren't the best.

bluss:
Not disagreeing about the plan for the future, but fixing the num-macros nightly issue seems like the reasonable first thing to do right now.

bluss:
#229 is the fix

hauleth:
So in sake of future changes:

- this PR rename `#[derive(NumFromPrimitive)]`  to more natural `#[derive(FromPrimitive)]`
- user should use `#![feature(rustc_macro)]` instead of `#![feature(custom_derive, plugin)]`
- user should use `#[macro_use] extern crate num_macros;` instead of `#![plugin(num_macros)]`

In future it should also be allowed to use `#[macro_use] extern crate num;` and also be able to use custom derives.

cuviper:
BTW, how did you get that "code broken after change" list?  Is crater available to us now?

hauleth:
@cuviper no, just GitHub search and copy&paste 😄

dtolnay:
I fixed the syn enum parsing issue in 0.6.0. Sorry for the slow response - I was away without internet access for a week.

bluss:
[cargo-crusader ](https://github.com/brson/cargo-crusader)was kind of a bare bones "crater" for rdeps, but it errors when I try to use it nowadays. Not sure if there was a crates.io api change or what happened.

macros/src/lib.rs
```diff
@@ -8,194 +8,59 @@
...
-                            "`FromPrimitive` cannot be derived for enums with no variants");
-                return cx.expr_fail(trait_span, InternedString::new(""));
+    let mut idx = 0;
+    let variants: Vec<_> = variants.iter()
+        .map(|variant| {
+            let ident = &variant.ident;
+            if let Some(val) = variant.discriminant {
+                idx = val.value;
+            }
+            let tt = quote!(#idx => Some(#name::#ident));
```
cuviper: Remember that enums can have data too, but `FromPrimitive` only makes sense for integral enums.  It looks like that's where `Variant.data` is `VariantData::Unit`.  If any variant is something else, we need to reject the whole thing.  

For instance, try adding something like `Other(u8, u8, u8)` to one of the `Color` tests.  It does get an error, but it's a very confusing one:

```
error[E0308]: mismatched types
  --> tests/trivial.rs:18:1
   |
18 | enum Color {
   | ^ expected enum `Color`, found fn item
   |
   = note: expected type `Color`
   = note:    found type `fn(u8, u8, u8) -> Color {Color::Other}`
```

The derivation tried to output `Some(Color::Other)` here, but that's actually a type constructor, not a value in itself, which is why the error is talking about a `fn`.  Similarly if it were `Other{ r:u8, g:u8, b:u8 }` you get the error `struct called like a function`.  We should catch this up front to identify the root problem.

We could use [compiletest-rs](https://github.com/laumann/compiletest-rs) to have some negative tests.

macros/Cargo.toml
```diff
@@ -1,17 +1,26 @@
...
+name = "num-macros"
+repository = "https://github.com/rust-num/num"
+version = "0.1.33"
 
-[lib]
-name = "num_macros"
-plugin = true
+[dependencies]
+quote = "0.1.3"
+syn = "0.6.0"
```
cuviper: There's even 0.7 now.  Don't know if it matters to us, but we probably should try to keep up.

dtolnay: Don't bother trying to keep up for now. I am sprinting to get serde codegen working with syn (https://github.com/serde-rs/serde/pull/548) so there may be another break coming up. When serde is done I will make a PR here bumping the version and fixing anything I broke (although I doubt there would be anything affecting the code in this PR). Also starting with the next release I will start doing proper release notes explaining the changes.

cuviper:
OK, thanks for telling us!

macros/Cargo.toml
```diff
@@ -1,17 +1,26 @@
...
-num = { path = "..", version = "0.1" }
+
+[dev-dependencies.num]
+path = ".."
+version = "0.1"
+
+[lib]
+crate-type = ["rustc-macro"]
+name = "num_macros"
+rustc-macro = true
```
dtolnay: From what I have seen in serde, this line is sufficient. You don't need `crate-type = ["rustc-macro"]` and `#![crate_type = "rustc-macro"]`. Is there a reason to have those?

macros/src/lib.rs
hauleth: Oh yeah, I have forgotten about that.

macros/Cargo.toml
hauleth: None, but it isn't harmful in any way and this helps with visibility (you do not need to check `Cargo.toml` for crate type, it is already there).

macros/src/lib.rs
hauleth: Added better error message and `compiletest-rs` crate to test against invalid code.

hauleth:
@cuviper all done.

dtolnay:
I haven't reviewed carefully but I just want to say what a massive improvement in readability of lib.rs, nicely done. Once this merges I will link to this from the `syn` and `quote` readmes as an example of how to do things right.

cuviper:
You didn't answer my question about the "Code broken after change" status, but I suppose they must still be broken since you changed the derivation name.  Let's back that change out and save it for 0.2, so we can merge the syn rewrite now.

cuviper:
Or if you *really* want the new name now, add a derivation by the old name too for compatibility.

hauleth:
@cuviper it is and will be broken even if I keep the name as "requiring" syntax has changed.

## Ex.

### Old

```rust
#![features(plugin, custom_derive)]
#![plugin(num_macros)]

extern crate num;

#[derive(FromPrimitive)]
enum Foo { Bar }

// ...
```

### New

```rust
#![features(rustc_macro)]

#[use_macros] extern crate num_macros;
extern crate num;

#[derive(FromPrimitive)]
enum Foo { Bar }
```

So either way it is still breaking change.

dtolnay:
For serde we decided to keep "serde_macros" as the feature(plugin) one and use "serde_derive" as the feature(rustc_macro) one. I think "derive" is a lot more evocative in terms of how the crate is used, and "macros" just happens to be an implementation detail of Rust.

cuviper:
OK, I'd be fine with forking `num_macros` into a `syn`-based `num_derive` crate, and leave the former alone.  That would also make it easier to show side-by-side for your `syn` case study. :)

hauleth:
Ok. I have moved it to new crate `num-derived` and restored old `num-macros`.

cuviper:
@homu r+ b7e6407

PR #229, opened 2016-09-18
Fix num-macros for nightly

bluss:
Now compatible with rustc 1.13.0-nightly (32571c05c 2016-09-17)


bluss:
Syntax extension breaks are usually super simple things like this. For now it's sensible to just not support deriving on unions (it would be a new feature, not suitable for a fix PR).

hauleth:
I was just working on that. But thanks 👍

hauleth:
@homu r+

bluss:
thanks!

PR #231, opened 2016-09-18
Implementing epsilon function to retrieve EPSILON constant

AtheMathmo:
Hey!

This PR exposes a new `epsilon` function in the `Float` trait so that users can access the `EPSILON` constant on the float types. I figured as this was such a minimal change it was easier to get a PR in offering the change then write up an issue.

For me this is a valuable addition. When writing linear algebra or other optimization routines with generic floats we often want to check if some stopping criteria is reached, often something like: `(a - b).abs() < eps`. Having access to a standard _small value_ would make this a little easier.

### DIFF
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -2,6 +2,9 @@ use std::mem;
 use std::ops::Neg;
 use std::num::FpCategory;
 
+// Used for default implementation of `epsilon`
+use std::f32;
+
 use {Num, NumCast};
 
 // FIXME: these doctests aren't actually helpful, because they're using and
@@ -99,7 +102,14 @@ pub trait Float
     ///
     /// assert_eq!(x, f64::EPSILON);
     /// ```
-    fn epsilon() -> Self;
+    ///
+    /// # Panics
+    ///
+    /// The default implementation will panic if `f32::EPSILON` cannot
+    /// be cast to `Self`.
+    fn epsilon() -> Self {
+        Self::from(f32::EPSILON).expect("Unable to cast from f32::EPSILON")
+    }
 
     /// Returns the largest finite value that this type can represent.
     ///
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -89,6 +89,18 @@ pub trait Float
     /// ```
     fn min_positive_value() -> Self;
 
+    /// Returns epsilon, a small positive value.
+    ///
+    /// ```
+    /// use num_traits::Float;
+    /// use std::f64;
+    ///
+    /// let x: f64 = Float::epsilon();
+    ///
+    /// assert_eq!(x, f64::EPSILON);
+    /// ```
+    fn epsilon() -> Self;
+
     /// Returns the largest finite value that this type can represent.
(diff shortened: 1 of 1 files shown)


cuviper:
It's a breaking change to add new methods to a trait without a default
implementation.

Note the f32/f64 values are just 2^-23 and 2^-52, directly related to the
mantissa size. We don't know the mantissa for generic floats either, but
perhaps there's some other way to estimate epsilon.  It doesn't have to be
fast or pretty, just best effort to avoid outright breaking.

(Or we can wait for the mass breaking change 0.2 that we keep talking
about...)

AtheMathmo:
Thank you for your comments! I didn't realize this was a breaking change - sorry about that!

I'm assuming that to avoid breakage you are implying we could provide a default implementation? I'm not too sure how this would work (without bringing in FromPrimitive or something - which is also a breaking change?). If you have any suggestions I'm happy to try and implement them or you can feel free to push directly to this branch.

cuviper:
Right, it avoids the break if you have a default impl.  Float doesn't have FromPrimitive, but it does have NumCast which can probably help. (I always thought that was a weirdly defined trait.)

AtheMathmo:
So perhaps we can provide a default implementation like this:

```rust
fn epsilon() -> T {
    T::from(f32::EPSILON)
}
```

And we leave the overrides for f32 and f64 in the macro? This way we have the correct implementations for f32 and f64 with a sensible (enough?) default for users implementing `Float`. If that is acceptable I'm happy to make this change.

cuviper:
Yeah, that's probably Good Enough -- it's hard to be confident that any other heuristic would truly be better.  It's `Self` instead of `T`, and you get an `Option` so you should `expect("some panic message")`.

AtheMathmo:
I've implemented the default as above. If you have a better panic message let me know!

(I also added a panic header in the docs - I can remove it if you prefer)

cuviper:
Looks good, thanks!  When we do make breaking changes, we might drop this default impl to make it mandatory for implementors.

@homu r=hauleth

AtheMathmo:
> Looks good, thanks! When we do make breaking changes, we might drop this default impl to make it mandatory for implementors.

Great, thank you!

Would you like me to open an issue to track the removal of the default impl with the next minor version bump? (I agree that it is probably best to remove it)

cuviper:
Sure, a tracking issue would be fine.

eddyb:
FWIW this broke [`glm`](https://tools.taskcluster.net/task-inspector/#ytvixlBAQzCMhtWXQCM3fA/0) and [`simple_gaussian`](https://tools.taskcluster.net/task-inspector/#WJwHtMdHQmikycmFEZGd3Q/0) (click `Logs` to see the output) due to ambiguity.

cuviper:
That's unfortunate, but allowable as a minor change under RFC#1105, thanks to the default impl.  It sure would be nice to have Crater access ourselves though.  I'll file heads-up bugs on those packages.

eddyb:
cc @brson for the crater thing.

PR #238, opened 2016-10-24
traits: Remove pattern in trait's method signature

bluss:
This use of `mut` was nonsensical for now, but is proposed to be
disallowed by rustc in the future (as a bugfix).

See rust-lang/rust/pull/37378 for context.


cuviper:
OK, thanks!

@homu r+ 96e96ef

PR #239, opened 2016-11-01
Add doc(html_root_url) and other doc attrs to each crate

bluss:
Also update to use https instead of http. This avoids mixed content
degradation on docs.rs.

The doc root URLs are correct as they are, the URL does not include the
crate name itself.

### DIFF
--- a/bigint/src/lib.rs
+++ b/bigint/src/lib.rs
@@ -72,7 +72,7 @@
 #![doc(html_logo_url = "https://rust-num.github.io/num/rust-logo-128x128-blk-v2.png",
        html_favicon_url = "https://rust-num.github.io/num/favicon.ico",
        html_root_url = "https://rust-num.github.io/num/",
-       html_playground_url = "https://play.rust-lang.org/")]
+       html_playground_url = "http://play.integer32.com/")]
 
 #[cfg(any(feature = "rand", test))]
 extern crate rand;
--- a/bigint/src/lib.rs
+++ b/bigint/src/lib.rs
@@ -69,6 +69,10 @@
 //! # fn main() {
 //! # }
 //! ```
+#![doc(html_logo_url = "https://rust-num.github.io/num/rust-logo-128x128-blk-v2.png",
+       html_favicon_url = "https://rust-num.github.io/num/favicon.ico",
+       html_root_url = "https://rust-num.github.io/num/",
+       html_playground_url = "https://play.rust-lang.org/")]
 
 #[cfg(any(feature = "rand", test))]
 extern crate rand;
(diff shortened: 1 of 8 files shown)


bluss:
Trying the integer32 playground. Some of the examples don't compile over there, because they are doc-included from one crate to another.

One example is a bigint example (BigInt::from_bytes_be) which produces [this playground gist](http://play.integer32.com/?gist=5ddee2c831fdcb16818fcf7c2256b485&version=stable)

bluss:
That's more or less a rustdoc bug and of course it won't compile on any playground.

bluss:
Main motivation for this  PR is so that `Zero` is correctly linked in ndarray's docs (without generating & hosting num docs).

After an eventual docs.rs migration, it's relatively easy to pin the doc URLs on a minor version like "0.1" so that cross crate URLs are somewhat versioned.

cuviper:
Hmm, that's interesting.  In an ideal world, those doc examples would adapt their `use` statements depending on which crate they're shown from, but I have no idea how to accomplish that.

cuviper:
My only complaint about docs.rs is that it doesn't link the crate imports very well.
e.g. this https://docs.rs/num/0.1.36/num/bigint/index.html
vs. this http://rust-num.github.io/num/num/bigint/index.html

But maybe it would look better in general if we used explicit imports there instead of a wildcard.

bluss:
Wouldn't it link that if the crates in question set up their doc_root_url?

cuviper:
I don't know -- if so, great!

bluss:
Local test says yes. That will result in docs.rs's crate cross links pointing to rust-num.github.io for now.

bluss:
By the way, you can control item inclusion in docs using #[doc(inline)] or #[doc(no_inline)] on glob imports.

cuviper:
I just tried this locally, to no effect:

```rust
pub mod bigint {
    #[doc(inline)]
    pub use num_bigint::*;
}
```

bluss:
Aha. I've only ever used it for crate-internal imports, so I guess it doesn't work the way I assumed here.

briansmith:
@bluss, are these things that every single crate needs to do, or is this somehow specific to `num`? If it is a general problem, I'm wondering if people shouldn't just submit PRs to docs.rs or whatever to fix the default behaviors?

cuviper:
I suppose from docs.rs's perspective, it only sees the separate crates from crates.io, rather than the crate hierarchy present here in git.  And usually when you do a local `cargo doc` you get all dependency docs lumped together.

bluss:
@briansmith All the crates need to do it as it is now. In the new centralized world, you're right that docs.rs could solve this much better. Rustdoc would need an override that tells it that everyone's root url is on docs.rs.

cuviper:
Anyway, the url updates look good.  Thanks!

@homu r+ ff2a350

bluss:
Thank you

PR #242, opened 2016-11-27
Fix code duplication error

### LINKED ISSUE #240: Failure when using num_derive on latest nightly
To reproduce, run `cargo test` on the `num/derive` subdir, you will get lots of "a type named ... has already been defined in this module".

I can reproduce with `1.15.0-nightly (0ed951993 2016-11-14)`.

This change in behaviour was probably caused by https://github.com/rust-lang/rust/pull/37614, but its actually wanted behaviour.

Basically the proc macro isn't expected anymore to output the item itself, only the impl's.

hauleth:
Fixes #240

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -30,10 +30,7 @@ pub fn from_primitive(input: TokenStream) -> TokenStream {
 
     let variants = match ast.body {
         Enum(ref variants) => variants,
-        _ => {
-            panic!("`FromPrimitive` can be applied only to the enums, {} is not an enum",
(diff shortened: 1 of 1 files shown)


est31:
:+1: I confirm that this fixes #240. Thanks!

est31:
If possible, can there be a new release of `num_derive` after this is merged?

cuviper:
I'll publish a new `num-derive` release shortly.

cuviper:
See `num-derive` 0.1.38.

PR #244, opened 2016-12-01
Added trait From<T> for Ratio<T> where T: Clone + Integer

wictory:
Hi!

So I thought that this kind of trait would be useful when one wants to write code which should be generic over different types of numbers.


cuviper:
Sure, looks reasonable to me.  Could you add a quick test somewhere?  Perhaps in `test_test_constants`.

wictory:
Added.

wictory:
Should I rebase?

cuviper:
I guess you mean squash -- but no, I don't mind if the PR has multiple commits.

Nightly CI is broken right now due to rust-lang/cargo#3340, but that should be resolved when a new nightly comes out, so I'm just going to wait on merging for a clean CI run.

cuviper:
@homu r+ 4bdad38

PR #248, opened 2016-12-09
rational: Into<(T,T)> implementation

dnsl48:
Right now it appears impossible to get integers without cloning them. Converting into a pair can be a cheap and easy way to extract the data.

### DIFF
--- a/rational/src/lib.rs
+++ b/rational/src/lib.rs
@@ -268,6 +268,14 @@ impl<T> From<T> for Ratio<T> where T: Clone + Integer {
     }
 }
 
+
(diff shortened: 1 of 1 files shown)


cuviper:
OK, I suppose this makes sense to have.

For symmetry, can you also add the reverse, `From<(T, T)>`, just mapping to `Ratio::new`?

dnsl48:
Sounds reasonable =)
This is done.

cuviper:
Hmm, I wasn't sure if homu would pick up the comment from a change review.  Oh well, try again...

@homu r+ 9d7ab66

PR #256, opened 2017-01-22
Added traits for wrapping arithmetics.

### LINKED ISSUE #255: Add `Wrapping[...]` traits (similar to `Checked` traits)
I was about to file an issue that we need these traits because I couldn’t find them.

We should resemble the wording that is used in the stdlib which has a type named `Wrapping` and several `wrapping_[...]` methods.

Any thoughts about that?

nwin:
Added `Wrapping` traits for the most common operations. Similar to the already present `Checked` traits.

It was quite strait-forward. I am unsure if the Handling of `Shl` / `Shr` is good. Their pattern is unfortunately a bit different than for the elementar mathematical operations by always taking `u32` as the RHS argument.

I just copied the docs from the stdlib. Is this ok license-wise?

Closes #255

### DIFF
--- a/traits/src/ops/wrapping.rs
+++ b/traits/src/ops/wrapping.rs
@@ -1,0 +1,153 @@
+use std::ops::{Add, Sub, Mul, Div, Shl, Shr};
+
+macro_rules! wrapping_impl {
+    ($trait_name:ident, $method:ident, $t:ty) => {
+        impl $trait_name for $t {
+            #[inline]
+            fn $method(&self, v: &Self) -> Self {
+                <$t>::$method(*self, *v)
+            }
+        }
+    };
+    ($trait_name:ident, $method:ident, $t:ty, $rhs:ty) => {
+        impl $trait_name<$rhs> for $t {
+            #[inline]
+            fn $method(&self, v: &$rhs) -> Self {
+                <$t>::$method(*self, *v)
+            }
+        }
+    }
+}
+
+/// Performs addition that wrapps around on overflow.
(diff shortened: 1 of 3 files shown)


cuviper:
> I am unsure if the Handling of Shl / Shr is good. Their pattern is unfortunately a bit different than for the elementar mathematical operations by always taking u32 as the RHS argument.

Yeah, it's weird that all the primitive types' `wrapping_shr` take `u32`, while they implement `Shr<usize>`, even on their `Wrapping<T>`.  :man_shrugging:  But as you already found, we can't add these anyway until we move beyond Rust 1.0.

> I just copied the docs from the stdlib. Is this ok license-wise?

IANAL, but I think this is fine, as we use the same license.  Any reason to think it's not?

traits/src/ops/wrapping.rs
```diff
@@ -0,0 +1,77 @@
...
+        impl $trait_name<$rhs> for $t {
+            #[inline]
+            fn $method(&self, v: &$rhs) -> Self {
+                <$t>::$method(*self, *v)
+            }
+        }
+    }
+}
+
+/// Performs addition that wrapps around on overflow.
```
cuviper: *wraps

cuviper:
On the Travis CI failure, see #257.  If we end up bumping our effective Rust minimum, you may get access to those additional wrapping methods you wanted after all. :)

In the meantime, this looks good, but can you add some tests?  Doesn't have to be anything exhaustive, just something to exercise the traits a little, e.g. scenarios where they do and don't wrap around.

cuviper:
Thanks!

@homu r+

Issue #257, opened 2017-01-27
serde's -rc4 dependency breaks old cargo


cuviper:
See https://users.rust-lang.org/t/crates-io-compatibility-with-old-cargo/9127

It appears that since we have a dependency on serde, even optional, then Cargo < 0.9 (w/ Rust 1.8) can't parse serde-0.9.0's dev-dependency on serde-derive 0.9.0-rc4, specifically the "-rc4" part.  It doesn't matter that this version is out of our range, and from my tests even yanking won't solve anything.

So what should num do?

I think we're stuck, because anyone referencing num with old Cargo is seemingly going to have this transitive problem, even if we were to now remove our serde feature entirely.  I made a [test crate](https://crates.io/crates/test_yank_rc_dep) to try this out, and even though the latest version has no dependencies at all, and all others are yanked, it still can't be used by old Cargo.

If somebody has a working Cargo.lock already, they should be fine, hopefully.  But for such a person it doesn't matter what we do now.  It may be that such old users are extinct by now anyway.

So... shall we embrace the break and raise our minimum Rust to 1.8?  That's the least that's not broken right now, through no fault of our own.  There are also some nice things we'll gain access to, like the `std::ops::*Assign` traits that were stabilized in 1.8.

cuviper:
One casualty will be Ubuntu 16.04, which has rustc 1.7 and cargo 0.8 in `universe`.

bluss:
Interesting issue and this seems like the only reasonable move.

PR #258, opened 2017-01-31
Add doc example for `num_traits::Num::from_str_radix`.

frewsxcv:


cuviper:
Looks good, thanks.  CI failure is #257, not your fault, but I have to bypass @homu.

PR #262, opened 2017-02-07
Add clamp function

Xaeroxe23:
This PR adds a "clamp" function to num.  Oftentimes it is desirable to restrict a value to between a certain minimum and maximum, the clamp function provides a generic way to do this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,7 @@ pub use num_traits::{Num, Zero, One, Signed, Unsigned, Bounded,
                      one, zero, abs, abs_sub, signum,
                      Saturating, CheckedAdd, CheckedSub, CheckedMul, CheckedDiv,
                      PrimInt, Float, ToPrimitive, FromPrimitive, NumCast, cast,
-                     pow, checked_pow};
+                     pow, checked_pow, clamp};
 
 #[cfg(feature = "num-bigint")]
 pub mod bigint {
@@ -108,16 +108,4 @@ pub mod traits {
 #[cfg(feature = "num-rational")]
 pub mod rational {
     pub use num_rational::*;
-}
-
-/// A value bounded by a minimum and a maximum
-///
-/// If input is less than min then min is returned, if input is greater than max then max is
-/// returned.  Otherwise input is returned.
-#[inline]
-pub fn clamp<T: PartialOrd + Copy>(input: T, min: T, max: T) -> T {
-    debug_assert!(min < max, "min must be less than max");
-    if input <= min {min}
-    else if input >= max {max}
-    else {input}
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,6 +114,7 @@ pub mod rational {
 ///
 /// If input is less than min then min is returned, if input is greater than max then max is
 /// returned.  Otherwise input is returned.
+#[inline]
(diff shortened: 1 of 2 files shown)


Xaeroxe23:
The build fails before it even starts compiling num.  The build server is likely misconfigured.

cuviper:
The 1.0 failure is #257 -- nothing you can can fix here.

I'll give a review here in a little bit.

clarcharr:
Why does this require `Copy`? It doesn't seem necessary.

src/lib.rs
```diff
@@ -109,3 +109,15 @@ pub mod traits {
 pub mod rational {
     pub use num_rational::*;
 }
+
+/// A value bounded by a minimum and a maximum
+///
+/// If input is less than min then min is returned, if input is greater than max then max is
+/// returned.  Otherwise input is returned.
+#[inline]
+pub fn clamp<T: PartialOrd + Copy>(input: T, min: T, max: T) -> T {
```
cuviper: I agree with @clarcharr that this shouldn't need `Copy`.

src/lib.rs
```diff
@@ -109,3 +109,15 @@ pub mod traits {
...
+    debug_assert!(min < max, "min must be less than max");
```
cuviper: I think `min <= max` is fine, if we're going to assert anything.

src/lib.rs
```diff
@@ -109,3 +109,15 @@ pub mod traits {
...
+    if input <= min {min}
+    else if input >= max {max}
+    else {input}
```
cuviper: This formatting is non-standard -- please run it through rustfmt.

I think it would be slightly preferable to return `input` when it is equal to the boundaries.  This only matters for cases similar to stable sorting, where there may be internal data that's not part of the ordering.  So just compare `input < min` and `input > max`.

Xaeroxe23:
Yeah good point.  Copy made sense at the time but it's not really needed.

Requested changes have been made.

clarcharr:
Another thing is that it'd be best to match on `Ord::cmp` instead of doing ifs.

cuviper:
> Another thing is that it'd be best to match on Ord::cmp instead of doing ifs.

Why?  Then it will require `Ord` too.

Xaeroxe23:
The alternative is [partial_cmp](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#tymethod.partial_cmp) but I don't understand what benefit we gain from using that.

Xaeroxe23:
```Rust
match input.partial_cmp(min) {
    Some(Ordering::Less) => min,
    _ => 
        match input.partial_cmp(max) {
            Some(Ordering::Greater) => max
            _ => input
        },
}
```

This is my best attempt at implementing it using partial_cmp.  I don't really like it.

cuviper:
@clarcharr See rust-lang/rust#39538, which found in that case that binary comparisons perform better than a ternary `cmp`.  Unless you can justify your suggestion further, I'm inclined to merge this as-is.

Xaeroxe23:
The suggestion would make sense if it were possible to implement it using just one match statement, but because it is being compared to two values, not one, the suggestion doesn't really work.

Xaeroxe23:
So @cuviper how long are we supposed to wait for a reply?

cuviper:
Merged - thanks!

PR #263, opened 2017-02-10
Bump rustc to 1.8 and revamp all CI

cuviper:
This bumps the minimum rustc to 1.8.0, "fixing" #257.  I normally
consider this a breaking change, but we were already broken due to
external factors, for which I couldn't find a workaround.

This adds 1.15.0 to the CI matrix to build stable num-derive.  We still
need nightly to run its tests though, because of compiletest_rs, and
dev-dependencies can't be optional.

The testing scripts are moved from .travis/ to ci/, as they don't really
need to be hidden.  It's also now consolidated into one test_full.sh
which considers $TRAVIS_RUST_VERSION as needed.


hauleth:
@homu r+

cuviper:
Hold on, it failed because I wrote TRAVIS_BRANCH when I meant TRAVIS_RUST_VERSION.

cuviper:
@homu retry

cuviper:
@homu r=hauleth

PR #267, opened 2017-02-26
Add licenses to subcrates

### LINKED ISSUE #266: please include LICENSE files into archives of subcrates on crates.io
Apache-2.0 requires text to be presented along with sources. Having MIT text license would not hurt as well.

cuviper:
cc @ignatenkobrain
Fixes #266

### DIFF
--- a/bigint/Cargo.toml
+++ b/bigint/Cargo.toml
@@ -7,7 +7,7 @@ keywords = ["mathematics", "numerics"]
 license = "MIT/Apache-2.0"
 name = "num-bigint"
 repository = "https://github.com/rust-num/num"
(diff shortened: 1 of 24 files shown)


ignatenkobrain:
@cuviper wondering if simple symlink to directory above would do that as well...

cuviper:
Yes, but I worry about cross-platform compatibility.  I did try that, and `cargo package` dereferences the link as we'd want.  But it *could* cause problems if a Windows user referenced num.git for some reason.  It's not a big deal to just duplicate them, as they don't change, and git will deduplicate the blobs anyway.

PR #269, opened 2017-03-06
bigint: Create the parsing error better for nested `+`

### LINKED ISSUE #268: BigUint::from_str_radix panics with invalid string containing "+"
Test case:
```
use num_traits::Num;
num_bigint::BigUint::from_str_radix("0+2", 10);
```
will panic instead of returning an Err:
```
stack backtrace:
...
  10:     0x5629031d9e61 - <core::result::Result<T, E>>::unwrap_err::ha2ff9ebcf57c662c
                        at /checkout/src/libcore/result.rs:789
  11:     0x5629031e38db - <num_bigint::biguint::BigUint as num_traits::Num>::from_str_radix::h2b33fa225070b5b3
                        at .../num-bigint-0.1.36/src/biguint.rs:245
```
Cause is that `num` tries to create and `unwrap_err` a ParseIntError from `std` by calling `parse` on `s[i..]` where `i` is the index of the `+`. This will not error out if there are only valid digits after the `+`.

@Manishearth for the trophy case - found by using cargo-fuzz on serde-pickle

cuviper:
If a `+` is encountered in the middle of parsing a BigUint, this should
generate an `ParseIntError::InvalidDigit`.  Since we can't create that
directly, we get it by trying to parse a `u64` from this point, but of
course `+` is a perfectly valid prefix to a `u64`.

Now we include the previous character in the string passed to `u64`, so it
has proper parsing context to understand what's in error.

Fixes #268.

### DIFF
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -3,8 +3,7 @@ use std::ops::{Add, Div, Mul, Neg, Rem, Shl, Shr, Sub};
 use std::str::{self, FromStr};
 use std::fmt;
 use std::cmp::Ordering::{self, Less, Greater, Equal};
-use std::{f32, f64};
-use std::{u8, i64, u64};
+use std::{i64, u64};
 use std::ascii::AsciiExt;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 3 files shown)


cuviper:
@homu r+

PR #270, opened 2017-03-11
Implement binomial

vks:
The implementation avoids overflows and fractions.

### DIFF
--- a/integer/src/lib.rs
+++ b/integer/src/lib.rs
@@ -668,11 +668,8 @@ impl_integer_for_usize!(usize, test_integer_usize);
 /// Calculate r * a / b, avoiding overflows and fractions.
 fn multiply_and_divide<T: Integer + Clone>(r: T, a: T, b: T) -> T {
     // See http://blog.plover.com/math/choose-2.html for the idea.
-    //
-    // This depends on the fact that `b` must evenly divide `r*a`, as that's
-    // what lets you know that `b/gcd(r, b)` divides `a` evenly.
     let g = gcd(r.clone(), b.clone());
-    (r/g.clone()) * (a / (b/g))
+    (r/g.clone() * a) / (b/g)
 }
 
 /// Calculate the binomial coefficient.
--- a/integer/src/lib.rs
+++ b/integer/src/lib.rs
@@ -665,6 +665,36 @@ impl_integer_for_usize!(u32, test_integer_u32);
 impl_integer_for_usize!(u64, test_integer_u64);
 impl_integer_for_usize!(usize, test_integer_usize);
 
+/// Calculate r * a / b, avoiding overflows and fractions.
+fn multiply_and_divide<T: Integer + Clone>(r: T, a: T, b: T) -> T {
+    // See http://blog.plover.com/math/choose-2.html for the idea.
+    let g = gcd(r.clone(), b.clone());
+    (r/g.clone()) * (a / (b/g))
+}
+
+/// Calculate the binomial coefficient.
+pub fn binomial<T: Integer + Clone>(mut n: T, k: T) -> T {
+    // See http://blog.plover.com/math/choose.html for the idea.
+    if k > n {
+        return T::zero();
+    }
+    let two = T::one() + T::one();
+    if k > n.clone()/two {
+        return binomial(n.clone(), n - k);
+    }
+    let mut r = T::one();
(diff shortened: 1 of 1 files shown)


integer/src/lib.rs
```diff
@@ -665,6 +665,36 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+}
+
+/// Calculate the binomial coefficient.
+pub fn binomial<T: Integer + Clone>(mut n: T, k: T) -> T {
+    // See http://blog.plover.com/math/choose.html for the idea.
+    if k > n {
+        return T::zero();
+    }
+    let two = T::one() + T::one();
+    if k > n.clone()/two {
```
cuviper: It's probably cheaper in general to compare `k > n - k`, so you can avoid a division.  Primitive integers should optimize divide-by-two to a right shift, but for bigint division is much more expensive than subtraction.  Too bad that `Integer` doesn't include `Shr`.

integer/src/lib.rs
```diff
@@ -692,3 +722,48 @@ fn test_lcm_overflow() {
...
 }
+
+#[test]
+fn test_binomial() {
+    macro_rules! check {
+        ($t:ty, $x:expr, $y:expr, $r:expr) => { {
+            let x: $t = $x;
+            let y: $t = $y;
+            let expected: $t = $r;
+            assert_eq!(binomial(x, y), expected);
```
cuviper: You should check `binomial(x, x-y)` too, and perhaps some tests of `k > n`.

integer/src/lib.rs
```diff
@@ -665,6 +665,36 @@ impl_integer_for_usize!(u32, test_integer_u32);
 impl_integer_for_usize!(u64, test_integer_u64);
 impl_integer_for_usize!(usize, test_integer_usize);
 
+/// Calculate r * a / b, avoiding overflows and fractions.
+fn multiply_and_divide<T: Integer + Clone>(r: T, a: T, b: T) -> T {
+    // See http://blog.plover.com/math/choose-2.html for the idea.
+    let g = gcd(r.clone(), b.clone());
+    (r/g.clone()) * (a / (b/g))
```
cuviper: It's worth noting that this depends on the fact that `b` must evenly divide `r * a`, as that's what lets you know that `b/gcd(r,b)` divides `a` evenly.  Otherwise, you'd have to stick with the blog's original association `((r/g) · a)/(b/g)` to handle the truncation properly.

clarcharr:
For consistency, it might make sense to call this `binom`; a lot of maths libraries do this. Rust also shortens names of maths functions like `hypot` as well.

integer/src/lib.rs
vks: Done.

integer/src/lib.rs
vks: Done.

integer/src/lib.rs
vks: This was a by accident, I changed it to use the original formula.

vks:
> For consistency, it might make sense to call this binom; a lot of maths libraries do this. Rust also shortens names of maths functions like hypot as well.

I don't really care about the name. That being said, we might later want to add a more general function that calculates the multinomial coefficients, and I think `multinom` would be an awkward name.

> Also, would it make sense to have a factorial function too?

Not sure how useful it would be. It does not make much sense for non-bigints, because it grows so quickly.

> I think there would be room for a checked version of this too, as it's really easy for binomials to generate huge values and overflow. A checked version might even be a little faster, as its multiply_and_divide could try r * a directly first, before relying on the more expensive gcd.

This could be done in a later pull request. There might be room for an iterator over the binomial coefficients as well. I also have an implementation for the calculation of the multinomial coefficients (they reduce to a product of binomial coefficients), they are more exotic though.

vks:
I can reproduce the Travis failure on nightly, but it is in a different crate.

vks:
The (unrelated) failure should be fixed on the current nightly.

integer/src/lib.rs
cuviper: Heh, I was giving you the benefit of the doubt.  I think it's actually *better* the way you had it first, because it's less likely to overflow.  However, I wasn't able to find an input where this actually matters, so no worries.

cuviper:
@homu r+

PR #271, opened 2017-03-17
Implement multinomial

vks:


integer/src/lib.rs
```diff
@@ -694,6 +696,22 @@ pub fn binomial<T: Integer + Clone>(mut n: T, k: T) -> T {
     r
 }
 
+/// Calculate the multinomial coefficient.
+///
+/// Panics if no integers are specified.
```
cuviper: Must we panic?  The formula is a simple product, so it seems natural to use the identity `1` for the empty case.

integer/src/lib.rs
```diff
@@ -694,6 +696,22 @@ pub fn binomial<T: Integer + Clone>(mut n: T, k: T) -> T {
...
+pub fn multinomial<T: Integer + Clone>(k: &[T]) -> T
+    where for<'a> T: Add<&'a T, Output = T>
+{
+    assert!(k.len() > 0, "need at least one integer");
+    let mut r = T::one();
+    let mut p = T::zero();
+    for i in k {
```
cuviper: As a minor optimization, we could skip zeros, for they have no effect on the result.  (basically `p+=0; r*=1`)

vks: The additional branch is actually slower for `u64`. It might be worth it for bigints.

Without skipping zeros:
```
test bench_multinomial ... bench:         276 ns/iter (+/- 3)
```

Skipping zeros:
```
test bench_multinomial ... bench:         307 ns/iter (+/- 11)
```

I used this benchmark:
```rust
#[bench]
fn bench_multinomial(b: &mut Bencher) {
    b.iter(|| multinomial(&[91u64, 6, 3]));
}
```

integer/src/lib.rs
vks: Fair enough, I changed it.

cuviper:
Looks good, thanks!

@homu r+ 3de345f

PR #272, opened 2017-03-17
Implement an iterator over the binomial coefficients

vks:
I'm not very happy with the excessive cloning, but to fix it the bounds on the type parameters would have to be excessive. We probably need something like [this](https://github.com/vks/discrete-log/blob/master/src/main.rs#L90) in `num-traits`.

### DIFF
--- a/integer/src/lib.rs
+++ b/integer/src/lib.rs
@@ -676,6 +676,23 @@ impl<T> IterBinomial<T>
     where T: Integer,
 {
     /// For a given n, iterate over all binomial coefficients binomial(n, k), for k=0...n.
+    ///
+    /// Note that this might overflow, depending on `T`. For the primitive
+    /// integer types, the following n are the largest ones for which there will
+    /// be no overflow:
+    ///
+    /// type | n
+    /// -----|---
+    /// u8   | 10
+    /// i8   |  9
+    /// u16  | 18
+    /// i16  | 17
+    /// u32  | 34
+    /// i32  | 33
+    /// u64  | 67
+    /// i64  | 66
+    ///
+    /// For larger n, `T` should be a bigint type.
     pub fn new(n: T) -> IterBinomial<T> {
         IterBinomial {
             k: T::zero(), a: T::one(), n: n
@@ -716,6 +733,23 @@ fn multiply_and_divide<T: Integer + Clone>(r: T, a: T, b: T) -> T {
 }
 
 /// Calculate the binomial coefficient.
+///
+/// Note that this might overflow, depending on `T`. For the primitive integer
+/// types, the following n are the largest ones possible such that there will
+/// be no overflow for any k:
+///
+/// type | n
+/// -----|---
+/// u8   | 10
+/// i8   |  9
+/// u16  | 18
+/// i16  | 17
+/// u32  | 34
+/// i32  | 33
+/// u64  | 67
+/// i64  | 66
+///
+/// For larger n, consider using a bigint type for `T`.
 pub fn binomial<T: Integer + Clone>(mut n: T, k: T) -> T {
     // See http://blog.plover.com/math/choose.html for the idea.
     if k > n {
--- a/integer/src/lib.rs
+++ b/integer/src/lib.rs
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
 impl_integer_for_usize!(u64, test_integer_u64);
 impl_integer_for_usize!(usize, test_integer_usize);
 
+/// An iterator over binomial coefficients.
+pub struct BinomialCoeff<T> {
+    a: T,
+    n: T,
+    k: T,
+}
+
+impl<T> BinomialCoeff<T>
+    where T: Integer,
+{
+    /// For a given n, iterate over all binomial coefficients ((k, n - k), binomial(n, k)).
+    pub fn new(n: T) -> BinomialCoeff<T> {
+        BinomialCoeff {
+            k: T::zero(), a: T::one(), n: n
+        }
+    }
+}
+
+impl<T> Iterator for BinomialCoeff<T>
+    where T: Integer + Clone,
+          for<'a> &'a T: std::cmp::PartialEq<&'a T>
+{
+    type Item = ((T, T), T);
+
+    fn next(&mut self) -> Option<((T, T), T)> {
+        if &self.k > &self.n {
+            return None;
+        }
+        self.a = if &self.k != &T::zero() {
+            (self.a.clone() * (self.n.clone() - self.k.clone() + T::one())) / self.k.clone()
+        } else {
+            T::one()
+        };
+        let r = Some(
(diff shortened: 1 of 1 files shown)


integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
 impl_integer_for_usize!(u64, test_integer_u64);
 impl_integer_for_usize!(usize, test_integer_usize);
 
+/// An iterator over binomial coefficients.
+pub struct BinomialCoeff<T> {
```
cuviper: bikeshed: I think I'd call this `IterBinomial`.

integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+    a: T,
+    n: T,
+    k: T,
+}
+
+impl<T> BinomialCoeff<T>
+    where T: Integer,
+{
+    /// For a given n, iterate over all binomial coefficients ((k, n - k), binomial(n, k)).
```
cuviper: It's odd to me that you include `(k, n - k)` in the output.  They're both redundant - `k` is available by converting from an `enumerate` index if you want it, or zipped with a `Range`, and `n - k` is easily derived from that.  I feel like this iterator should just stick to its interesting contribution, just the coefficients.

integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+    pub fn new(n: T) -> BinomialCoeff<T> {
+        BinomialCoeff {
+            k: T::zero(), a: T::one(), n: n
+        }
+    }
+}
+
+impl<T> Iterator for BinomialCoeff<T>
+    where T: Integer + Clone,
+          for<'a> &'a T: std::cmp::PartialEq<&'a T>
```
cuviper: I don't think you need this `PartialEq<&'a T>`.  We already get `PartialEq<T>` from `Integer`, and there are blanket implementations in the standard library for reference variants from there.

integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+{
+    type Item = ((T, T), T);
+
+    fn next(&mut self) -> Option<((T, T), T)> {
+        if &self.k > &self.n {
```
cuviper: Furthermore, you don't actually need references here.  Just `self.k > self.n` is fine, and won't consume the operands.

integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+            return None;
+        }
+        self.a = if &self.k != &T::zero() {
```
cuviper: This too doesn't need references, and it can even be just `!self.k.is_zero()`.

integer/src/lib.rs
```diff
@@ -665,6 +665,47 @@ impl_integer_for_usize!(u32, test_integer_u32);
...
+            (self.a.clone() * (self.n.clone() - self.k.clone() + T::one())) / self.k.clone()
```
cuviper: I'm surprised this doesn't need to use your `multiply_and_divide`, as you're basically walking the same series of operations that `binomial` does.

integer/src/lib.rs
```diff
@@ -723,6 +764,46 @@ fn test_lcm_overflow() {
 }
 
 #[test]
+fn test_binomial_coeff() {
```
cuviper: Perhaps this needs to use one of `binomial`'s inputs that tread near overflow conditions?

integer/src/lib.rs
vks: They are included for convenience. For example, when calculating `(x + y)**n`, the iterator would return the coefficients of `x**k * y**(n - k)`. I could live without them though, they are more important for the multinomial coefficients, where the order is not as obvious.

integer/src/lib.rs
vks: Will change.

integer/src/lib.rs
vks: Good point, I changed it.

integer/src/lib.rs
vks: Done.

integer/src/lib.rs
vks: It actually might, I did not check it very carefully for possible overflows. Will do.

vks:
I checked the maximal n for the different algorithms before the first overflow happens. The last column gives an upper bound (the largest n for which the *final result* would overflow).

| type | max. n iterator | max. n iterator with `multiply_and_divide` | max. n `binomial` | max. n theoretical |
| --- | --- | --- | --- | --- |
| u8 | 7 | 8 | 10 | 14 |
| u16 | 15 | 16 | 18 | 31 |
| u32 | 30 | 30 | 33 | 62 |
| u64 | 62 | 62 | 66 | 126 |

We are far below the theoretical upper bound and `binomial` works better than the iterator. Unfortunately using `multiply_and_divide` does not seem to improve the iterator much.

The algorithms are different, because they use a different recursion relation. If you just want to know one coefficient, you can use `binomial(n, k) = n / k * binomial(n - 1, k - 1)`. If you want to know all binomial coefficients with a given n, you have to use `binomial(n, k) = (n + 1 - k) / k * binomial(n, k - 1)` instead.

Do you think it is still worth it to use `mutliply_and_divide`?

vks:
I wonder whether an implementation based on fractions makes sense. We have to calculate the gcd every iteration anyway and this should probably come closer to the theoretical upper bound.

vks:
Actually nevermind, using fraction will just overflow earlier.

cuviper:
One of the things we discussed before was for that `r*a/b` and `g=gdc(r,b)`, whether we should use `(r/g * a) / (b/g)` or `(r/g) * (a / (b/g))`.  You had the latter at first, and changed to the former when I commented on it.  But I think the latter is less likely to overflow, and is still correct as long as `b` divides `r*a` evenly, which it should here.  Can you see if that has any effect on your observed limits?

vks:
Using the modified `multipliy_and_divide`, `binomial` looks only slightly better, but more importantly `IterBinomial` becomes as well performing as `binomial`:

| type | `IterBinomial` | `binomial` |
| --- | --- | --- |
| `u8` | 10 | 10 |
| `u16` | 18 | 18 |
| `u32` | 34 | 34 |
| `u64` | 67 | 67 |

I'll push a corresponding change. Should I add such a table to the documentation?

cuviper:
Sure, having a bit of discussion about overflow and showing a table like that sounds good.

vks:
I added some documentation on overflow. This is now ready from my side.

cuviper:
Thanks!

@homu r+ 0fc6cb1

vks:
Thanks for the comprehensive review, it really improved the code!

PR #274, opened 2017-04-02
Implement OpAssign for Complex

bluss:
Design choice: Use by-value forms of OpAssign and use Clone to forward the reference arguments. That matches what the other traits do (for example Add).

Design choice: The complex += complex operation needs Add, not just AddAssign. The complex += real operation can use just AddAssign. You could just use Add for both, not sure which way is preferable.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -516,6 +516,88 @@ impl<T: Clone + Num> Div<Complex<T>> for Complex<T> {
     }
 }
 
+// Op Assign
+
+mod opassign {
+    use std::ops::{AddAssign, SubAssign, MulAssign, DivAssign};
+
+    use traits::Num;
+
+    use Complex;
+
+    impl<T: Clone + Num> AddAssign for Complex<T> {
+        fn add_assign(&mut self, other: Complex<T>) {
+            *self = self.clone() + other;
+        }
+    }
+
+    impl<T: Clone + Num> SubAssign for Complex<T> {
+        fn sub_assign(&mut self, other: Complex<T>) {
+            *self = self.clone() - other;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


bluss:
#186's revival.

cuviper:
Are you opposed to now adding a `NumAssign` trait, as discussed before?  Then each of these should require a full `T: Clone + NumAssign`.

bluss:
NumAssign sounds great.

cuviper:
`NumAssign` is in #283.

cuviper:
Can you update this to use `NumAssign`?

atilag:
Would love to see this PR landing. Thanks for the effort!!

atilag:
So I tried this PR, but seems like I can't multiply two Matrices, like:
```
let temp1 = DMatrix::<Complex<f64>>::identity(5usize, 5usize);
let temp2 = DMatrix::<Complex<f64>>::identity(5usize, 5usize);
println!(" Multiplication: {}", temp1 * temp2 );
```
Compiler error:
```
error[E0308]: mismatched types
  --> src/lib.rs:56:36
   |
56 |     println!("Multiplication: {:?}", temp1 * temp2);
   |                                    ^^^^^ expected struct `num::Complex`, found struct `na::Matrix`
   |
   = note: expected type `num::Complex<f64>`
              found type `na::Matrix<num::Complex<f64>, na::Dynamic, na::Dynamic, na::MatrixVec<num::Complex<f64>, na::Dynamic, na::Dynamic>>`
```

cuviper:
Rebased and applied `NumAssign`.  Thanks!

bors r+

bluss:
Thank you -- sorry I was absent.

PR #275, opened 2017-04-02
Complex: Use repr(C) and add documentation for what it means

### LINKED ISSUE #79: Complex struct should be #repr(C)
The memory layout is probably already identical to C's complex type or `std::complex<T>`, but it'd be nice to make it explicit. This would help with C interop, and things like real-only FFTs that need to cast real-only vectors into vectors of alternating real/imaginary.

bluss:
Here's an ambitious proposal that puts currently practiced ffi usage of `Complex<f32/f64>` on sound footing.

Fixes  #79

Current users appear to be:

- https://crates.io/crates/blas
  + Their use is not about Complex<f64> in an extern function's signature, but it is explicitly using that it is memory layout compatible.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -33,8 +33,33 @@ use traits::{Zero, One, Num, Float};
 // probably doesn't map to C's _Complex correctly.
 
 /// A complex number in Cartesian form.
+///
+/// ## Representation and Foreign Function Interface Compatibility
+///
+/// `Complex<T>` is memory layout compatible with an array `[T; 2]`.
+///
+/// Note that `Complex<F>` where F is a floating point type is **only** memory
+/// layout compatible with C's complex types, **not** necessarily calling
+/// convention compatible.  This means that for FFI you can only pass
+/// `Complex<F>` behind a pointer, not as a value.
+///
+/// ## Examples
+///
+/// Example of extern function declaration.
+///
+/// ```
+/// use num_complex::Complex;
+/// use std::os::raw::c_int;
+///
+/// extern "C" {
+///     fn zaxpy_(n: *const c_int, alpha: *const Complex<f64>,
+///               x: *const Complex<f64>, incx: *const c_int,
+///               y: *mut Complex<f64>, incy: *const c_int);
+/// }
+/// ```
 #[derive(PartialEq, Eq, Copy, Clone, Hash, Debug, Default)]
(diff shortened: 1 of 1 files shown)


bluss:
cc @IvanUkhov

complex/src/lib.rs
```diff
@@ -33,8 +33,34 @@ use traits::{Zero, One, Num, Float};
...
+///
+/// ## Representation and Foreign Function Interface Compatibility
+///
+/// `Complex<T>` is currently marked `#[repr(C)]` which means that it is *memory
+/// layout* compatible with an array `[T; 2]` as well as C's complex floating
+/// point types.
+///
+/// Note that `Complex<F>` where F is a floating point type is **only** memory
+/// layout, not necessarily **calling convention** compatible. This means for
+/// ffi you can only pass `Complex<F>` behind a pointer, not as a value.
```
cuviper: I appreciate keeping this simple, that we won't make any guarantees about calling convention.  I had trouble parsing it though -- the gap from "memory layout" to "compatible" is a bit awkward.  Alternates:

- add a comma: "is only memory layout, not necessarily calling convention, compatible."
- repeated: "is only memory layout compatible, not necessarily calling convention compatible."
- reordered: "is only compatible in memory layout, not necessarily calling convention."

I think I prefer it reordered.

Also, I'd apply symmetry to the bold style, either "**only**"/"**not**" or "**memory layout**"/"**calling convention**".

These are nits, but we're trying to clarify an important point here.

bluss:
It's hard for us to do this minimally because it only concerns `Complex<f32>, Complex<f64>`, and we can't scope the `repr(C)` like that. We want to gracefully add calling convention compatibility when that becomes possible.

With this in mind, maybe not mention `repr(C)` at all?

cuviper:
I don't care about mentioning `repr(C)` in particular, but shouldn't we still say that it's *C* layout compatibility we're talking about here, both for arrays and `_Complex`?

bluss:
Heh, good point. I'm not sure how to formulate it.

bluss:
I can squash the branch when we're done editing.

cuviper:
Ok, it works for me.  Squash if you like, then I'll signal @homu.

cuviper:
@homu r+ 0cf3419

bluss:
Thanks!

Issue #278, opened 2017-04-20
Shouldn't Num be also implemented for std::num::Wrapping<T> ?


yoanlcq:
That would be convenient - I'd like to be able to pass a `std::num::Wrapping<T>` to anything that accepts a `Num` type.  
  
Part of the discussion on [https://github.com/rust-num/num/pull/175](https://github.com/rust-num/num/pull/175) went over this, so I'd like to share my current use case for such a thing :  
I've experimented with a custom "marker" trait named `Deterministic` which is implemented on types on which basic operations are guaranteed to be consistent across platforms, build settings, etc.  
- `std::num::Wrapping<T>` implements `Deterministic`;
- fixed-point numbers are good candidates;
- primitive integer types aren't, because they might cause panics when they overflow;
- floating-point number types certainly aren't.

With this, I'd like to write functions which only accept types that are both `Num` and `Deterministic` (for some kind of reproducible game state data), and I can't pass `Wrapping<T>`s on them - the only workarounds are either to redefine my own Wrapping<T> type and have it implement Num, or define my own Num trait and... well, in short, nothing very appealing.
This "problem" is easy to solve from within this crate, but it seems more like a matter of agreeing on what the `Num` trait actually means. As far as I'm concerned, this crate should make `Wrapping<T>` implement `Num`.  
Thoughts ?

cuviper:
It still feels a little weird to me for `Num` to permit silent overflow behavior, but I guess it's up to the user of they want that. Plus it's not something we can forbid anyway, as one can always write their own wrapping type like you mentioned.  So, OK.  Care to send a PR?

yoanlcq:
Yup! See [https://github.com/rust-num/num/pull/279](https://github.com/rust-num/num/pull/279).
To add to the discussion, `Num` _does_ already permit silent overflow behaviour to some extent, because (if I'm not mistaken) primitive integer types are allowed to overflow on release builds. My 2 cents.

cuviper:
Fixed by #279.

PR #279, opened 2017-04-21
Implemented Zero, One and Num for std::num::Wrapping<T>

yoanlcq:
Attempts to fix issue #278

### DIFF
--- a/traits/src/identities.rs
+++ b/traits/src/identities.rs
@@ -1,4 +1,5 @@
 use std::ops::{Add, Mul};
+use std::num::Wrapping;
 
 /// Defines an additive identity element for `Self`.
 pub trait Zero: Sized + Add<Self, Output = Self> {
@@ -50,6 +51,16 @@ zero_impl!(i64,   0i64);
 zero_impl!(f32, 0.0f32);
 zero_impl!(f64, 0.0f64);
 
+impl<T: Zero + PartialEq> Zero for Wrapping<T> where Self: Add<Output=Self> {
+    fn is_zero(&self) -> bool {
+        self.0 == T::zero()
+    }
+    fn zero() -> Self {
+        Wrapping(T::zero())
+    }
+}
+
+
 /// Defines a multiplicative identity element for `Self`.
 pub trait One: Sized + Mul<Self, Output = Self> {
     /// Returns the multiplicative identity element of `Self`, `1`.
@@ -94,6 +105,11 @@ one_impl!(i64,   1i64);
 one_impl!(f32, 1.0f32);
 one_impl!(f64, 1.0f64);
 
+impl<T: One> One for Wrapping<T> where Self: Add<Output=Self> + Mul<Output=Self> {
+    fn one() -> Self {
+        Wrapping(T::one())
+    }
+}
 
 // Some helper functions provided for backwards compatibility.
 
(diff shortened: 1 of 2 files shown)


traits/src/lib.rs
```diff
@@ -243,9 +256,9 @@ float_trait_impl!(Num for f32 f64);
 
 /// A value bounded by a minimum and a maximum
 ///
-///  If input is less than min then this returns min.
```
yoanlcq: I have no idea what caused these changes - didn't do these on purpose anyway.

traits/src/identities.rs
```diff
@@ -50,6 +51,16 @@ zero_impl!(i64,   0i64);
 zero_impl!(f32, 0.0f32);
 zero_impl!(f64, 0.0f64);
 
+impl<T: Zero + PartialEq> Zero for Wrapping<T> where Wrapping<T>: Add<Output=Wrapping<T>> {
```
cuviper: Hmm.  I was worried that a generic `impl` would be a breaking change, in case someone had already implemented it for their own `Foo` and `Wrapping<Foo>`, but it looks like they couldn't do that:

```
    |
115 | impl Zero for Wrapping<Foo> {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ impl doesn't use types inside crate
    |
    = note: the impl does not reference any types defined in this crate
    = note: define and implement a trait or new type instead
```

... even though `Foo` is defined on the line right before.  I always forget exactly how coherence works.  I think you _can_ have this overlap with users if the custom type were a trait parameter instead.

Anyway, looks like this is OK.

traits/src/identities.rs
```diff
@@ -50,6 +51,16 @@ zero_impl!(i64,   0i64);
...
+    fn is_zero(&self) -> bool {
+        self.0 == T::zero()
```
cuviper: Should use `self.0.is_zero()` here.

traits/src/lib.rs
```diff
@@ -74,6 +75,18 @@ macro_rules! int_trait_impl {
...
 int_trait_impl!(Num for usize u8 u16 u32 u64 isize i8 i16 i32 i64);
 
+impl<T: Num> Num for Wrapping<T>
+    where Wrapping<T>: Zero + One
+        + Add<Output = Wrapping<T>> + Sub<Output = Wrapping<T>>
+        + Mul<Output = Wrapping<T>> + Div<Output = Wrapping<T>> + Rem<Output = Wrapping<T>>
+{
+    type FromStrRadixErr = T::FromStrRadixErr;
+    fn from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr> {
+        T::from_str_radix(str, radix).map(|x| Wrapping(x))
```
cuviper: I think this can just be `map(Wrapping)`, but it doesn't make much difference.

traits/src/lib.rs
cuviper: I guess your editor is set to strip trailing whitespace.  That's OK.

traits/src/lib.rs
yoanlcq: Wow, that's a pretty nice trick!

traits/src/identities.rs
```diff
@@ -94,6 +105,11 @@ one_impl!(i64,   1i64);
 one_impl!(f32, 1.0f32);
 one_impl!(f64, 1.0f64);
 
+impl<T: One> One for Wrapping<T> where Wrapping<T>: Add<Output=Wrapping<T>> + Mul<Output=Wrapping<T>> {
```
cuviper: I think this only needs `Mul`, not `Add`.

yoanlcq: I vaguely remember that I had to put it for some reason, but now it should indeed be removed.

cuviper:
Thanks!

@homu r+ 2172a93

Issue #94, opened 2015-05-13
Numeric types should probably support "reference overloads" for operators.


sellibitze:
Writing generic code over numeric types is hard in Rust right now. One has to add additional bounds like Clone, Copy or a whole bunch of `for<'a> &'a T: for <'b> Add<&'b T, Output=T>` bounds (for various combinations of references, non-references and operators) to support "non-consuming" versions of arithmetic operators.

What do you guys and gals think about adding those "non-consuming operators" to the `Num` trait? I added these kinds of operator implementations for the built-in types in the standard library for that reason in consultation with @aturon . Addings these requirements to the `Num` trait would make the life of a numeric type provider a little harder but we still have macros that ease the pain. It would be a great benefit for generic numeric code in that "T: Num" is much less noisy than "T: Num + for<'b> Add<&'b T, Output=T> + …` etc.

cuviper:
I hesitate to expand to requirements on `Num`, but maybe a new `RefNum`?

```rust
/// Numeric types with operators by reference
pub trait RefNum
where for<'b> Self: Num + Add<&'b Self, Output=Self> // ...
, for<'a, 'b> &'a Self: Add<Self, Output=Self> + Add<&'b Self, Output=Self> // ...
{ }

// Implement it automatically for qualifying types
impl<T> RefNum for T
where for<'b> T: Num + Add<&'b T, Output=T> // ...
, for<'a, 'b> &'a T: Add<T, Output=T> + Add<&'b T, Output=T> // ...
{ }

// example use:
fn foo<T: RefNum>() -> T { T::zero() + &T::one() }
```

But my lifetimes are missing something here -- 1.3 is ok, but 1.4 warns:

```
warning: the trait `for<'a> core::ops::Add<T>` is not implemented for the type `&'a T` [E0277]
fn foo<T: RefNum>() -> T { T::zero() + &T::one() }
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
help: run `rustc --explain E0277` to see a detailed explanation
note: this warning results from recent bug fixes and clarifications; it will become a HARD ERROR in the next release. See RFC 1214 for details.
fn foo<T: RefNum>() -> T { T::zero() + &T::one() }
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
note: required by `traits::RefNum`
[...]

PR #283, opened 2017-04-24
Add new traits for reference and assignment operators

cuviper:
There are two new "utility" traits covering the basic operators:
`Add`, `Sub`, `Mul`, `Div`, and `Rem`.

- `NumOps<Rhs, Output>`: operators with an arbitrary operand and output.
- `NumAssignOps<Rhs>`: assignment operators with an arbitrary operand.

Then the new collection of numeric traits are:

- `Num`: effectively unchanged, just taking operands by value.
- `NumRef`: `Num` adding reference operands on the right side.
- `RefNum`: `&T` operators, with either `T` or `&T` on the right side.
  - This does not specify `T: Num`, as rust-lang/rust#20671 means that
    could only add a constraint, without implying its presence for use.
- `NumAssign`: `Num` adding assignment operators by value.
- `NumAssignRef`: `NumAssign` adding reference assignment operators.
  - Nothing actually implements this yet!

Acknowledgement: this is roughly based on [@andersk's suggestion](https://github.com/rust-num/num/issues/94#issuecomment-269073071).

### DIFF
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -89,9 +89,9 @@ impl<T> NumRef for T where T: Num + for<'r> NumOps<&'r T> {}
 /// The trait for references which implement numeric operations, taking the
 /// second operand either by value or by reference.
 ///
-/// This is automatically implemented for all `&T` implementing the operators.
+/// This is automatically implemented for types which implement the operators.
 pub trait RefNum<Base>: NumOps<Base, Base> + for<'r> NumOps<&'r Base, Base> {}
-impl<'a, T> RefNum<T> for &'a T where &'a T: NumOps<T, T> + for<'r> NumOps<&'r T, T> {}
+impl<T, Base> RefNum<Base> for T where T: NumOps<Base, Base> + for<'r> NumOps<&'r Base, Base> {}
 
 /// The trait for types implementing numeric assignment operators (like `+=`).
 ///
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -15,6 +15,7 @@
        html_playground_url = "http://play.integer32.com/")]
 
 use std::ops::{Add, Sub, Mul, Div, Rem};
+use std::ops::{AddAssign, SubAssign, MulAssign, DivAssign, RemAssign};
 use std::num::Wrapping;
 
 pub use bounds::Bounded;
@@ -37,10 +38,9 @@ pub mod cast;
 pub mod int;
 pub mod pow;
 
-/// The base trait for numeric types
-pub trait Num: PartialEq + Zero + One
-    + Add<Output = Self> + Sub<Output = Self>
-    + Mul<Output = Self> + Div<Output = Self> + Rem<Output = Self>
+/// The base trait for numeric types, covering `0` and `1` values,
+/// comparisons, basic numeric operations, and string conversion.
+pub trait Num: PartialEq + Zero + One + NumOps
 {
     type FromStrRadixErr;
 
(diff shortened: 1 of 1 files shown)


traits/src/lib.rs
```diff
@@ -60,6 +60,72 @@ pub trait Num: PartialEq + Zero + One
...
+///
+/// This is automatically implemented for types which implement the operators.
+pub trait NumRef: Num + for<'r> NumOps<&'r Self> {}
+impl<T> NumRef for T where T: Num + for<'r> NumOps<&'r T> {}
+
+/// The trait for references which implement numeric operations, taking the
+/// second operand either by value or by reference.
+///
+/// This is automatically implemented for all `&T` implementing the operators.
+pub trait RefNum<Base>: NumOps<Base, Base> + for<'r> NumOps<&'r Base, Base> {}
```
cuviper: I don't like taking the `Base` type as a parameter, but I didn't find a better way to do this.  I tried adding a `Deref` constraint, but the compiler didn't seem to see that `<Self as Deref>::Target` always matches the types we want, especially when chaining as in the `check_refnum_ops` test.

traits/src/lib.rs
```diff
@@ -60,6 +60,72 @@ pub trait Num: PartialEq + Zero + One
...
+/// The trait for types implementing basic numeric operations
+///
+/// This is automatically implemented for types which implement the operators.
+pub trait NumOps<Rhs = Self, Output = Self>
+    : Add<Rhs, Output = Output>
+    + Sub<Rhs, Output = Output>
+    + Mul<Rhs, Output = Output>
+    + Div<Rhs, Output = Output>
+    + Rem<Rhs, Output = Output>
+{}
```
cuviper: I can't find a way to specify that an associated `Output` and all the individual ops' `Output` must be identical.  If you figure it out, let me know...

traits/src/lib.rs
```diff
@@ -60,6 +60,72 @@ pub trait Num: PartialEq + Zero + One
...
+/// This is automatically implemented for types which implement the operators.
+pub trait NumRef: Num + for<'r> NumOps<&'r Self> {}
+impl<T> NumRef for T where T: Num + for<'r> NumOps<&'r T> {}
+
+/// The trait for references which implement numeric operations, taking the
+/// second operand either by value or by reference.
+///
+/// This is automatically implemented for all `&T` implementing the operators.
+pub trait RefNum<Base>: NumOps<Base, Base> + for<'r> NumOps<&'r Base, Base> {}
+impl<'a, T> RefNum<T> for &'a T where &'a T: NumOps<T, T> + for<'r> NumOps<&'r T, T> {}
```
cuviper: In part, because I was hoping to find a way not to need the `Base` parameter at all.  I'm stuck on that though.

I also think it's a little weird that `Base = T` would be allowed, as I'm trying to specify a reference type here.  I was going for `NumRef` for values with numeric operators, and `RefNum` for references with numeric operators.  Maybe that distinction isn't so useful?  I'm not sure...

I probably could narrow it with `RefNum<Base>: Deref<Target = Base> + ...`

cuviper: The more I think about it, the more I like adding `Deref` in there, to lock down the intended relationship between `T` and `Base`.  If I can't get rid of `Base`, at least I can make it have only one possible answer -- `Base = <T as Deref>::Target`.

andersk: I can’t seem to make it work without the `Base` parameter either. This may be an inference bug. But unless `Base` can actually be avoided, such narrowing seems to add nothing to the present use case—in `for<'a> &'a T: RefNum<T>`, we don’t need to tell the compiler again that `&'a T` is a reference—and it might exclude other use cases we haven’t thought of.

cuviper: I'm trying to think if there are hypothetical situations where you'd want a clear distinction between `RefNum` and `NumRef`. Maybe for specialization? But I think you could use private traits with a `Deref` or similar differentiator. Dunno.

I'll play a little more, and as long I can't find any problem doing it your way, I'll switch it.

Thanks for your input!

cuviper:
OK, I switched to a more generic `impl` for `RefNum`.

cuviper:
I'll merge this in a day or two if there are no further comments...

cuviper:
@homu r+

cuviper:
@homu r+

Issue #282, opened 2017-04-24
f32/f64 to Ratio conversion


sdroege:
Is there any interest for having conversion from f32/f64 to Ratio? I've implemented that in my code, based on a continued fractions algorithm: https://github.com/sdroege/gst-plugin-rs/blob/d72f1f716b630bc31973fc559c5f5d82255965a1/gst-plugin/src/utils.rs#L53

The same algorithm is used in GStreamer since a decade or so, for handling framerates stored as a floating point number and similar things, and it generally gives good results, but it's of course not the only way of doing that and depending on the input values another algorithm might find "nicer" results.

If there's interest, I could prepare a pull request for moving that code over (and generalize it a bit).

cuviper:
There's already [`from_float`](http://rust-num.github.io/num/num_rational/struct.Ratio.html#method.from_float) on `Ratio<BigInt>` (aka `BigRational`), which is a bit easier since it can represent the full mantissa and exponent precisely.

I think it would be very reasonable to add the same method for `Rational32` and `Rational64`.  Even better if we could do it for generic `Ratio<T>`, but that seems harder to deal with generic boundary conditions, and we'd also want specialization to let the `BigRational` still use a direct conversion.

cuviper:
Hmm, there might be a downside (breaking change?) for type inference though, since right now you can write just `Ratio::from_float` and infer that it will be a `BigRational`.

sdroege:
We could make it a different function. ::approximate_from_float() or similar, as this will just give an approximation (should the error be returned?) and not an exact value like ::from_float() currently does.

cuviper:
How about implementing `FromPrimitive`?  That would get us `from_f32`, `from_f64`, and all the integer types too.

sdroege:
Seems like an option, yes. Has the disadvantage that it would be for primitive types only, and not for Float. Not sure if that matters :)

cuviper:
Fair point!  I think `FromPrimitive` is probably good enough, worthwhile on its own, and still leaves the hypothetical door open for something even more general.

sdroege:
Let's go with that then :) thanks

sdroege:
With some refactoring I got it generalized to
```rust
fn float_to_rational<T, F>(val: F, max_error: F, max_iterations: usize) -> Option<Ratio<T>>
    where T: Integer + Signed + Bounded + NumCast + Copy,
          F: Float + NumCast;
```

[...]

PR #285, opened 2017-04-28
rational: Implement approximation from floats and FromPrimitive for v…

sdroege:
…arious types

FromPrimitive is implemented for i8/16/32/64 and BigInt.

https://github.com/rust-num/num/issues/282

### DIFF
--- a/rational/src/lib.rs
+++ b/rational/src/lib.rs
@@ -36,7 +36,7 @@ use std::str::FromStr;
 use bigint::{BigInt, BigUint, Sign};
 
 use integer::Integer;
-use traits::{FromPrimitive, Float, PrimInt, Num, Signed, Zero, One};
+use traits::{FromPrimitive, Float, PrimInt, Num, Signed, Zero, One, Bounded, NumCast};
 
 /// Represents the ratio between 2 numbers.
 #[derive(Copy, Clone, Hash, Debug)]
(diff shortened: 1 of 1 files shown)


rational/src/lib.rs
```diff
@@ -668,6 +668,164 @@ impl RatioErrorKind {
...
+    fn from_f64(n: f64) -> Option<Self> {
+        Ratio::from_float(n)
+    }
+}
+
+macro_rules! from_primitive_integer {
+    ($typ:ty) => {
+        impl FromPrimitive for Ratio<$typ> {
+            fn from_i64(n: i64) -> Option<Self> {
+                <$typ as NumCast>::from(n).map(|n| Ratio::new(n, One::one()))
```
cuviper: Perhaps more directly as `$typ::from_i64(n).map(Ratio::from_integer)`?

rational/src/lib.rs
```diff
@@ -668,6 +668,164 @@ impl RatioErrorKind {
...
+                Self::approximate_float(n)
+            }
+        }
+    }
+}
+
+from_primitive_integer!(i8);
+from_primitive_integer!(i16);
+from_primitive_integer!(i32);
+from_primitive_integer!(i64);
```
cuviper: Might as well add `isize` at least, but why not any unsigned types?

rational/src/lib.rs
```diff
@@ -668,6 +668,164 @@ impl RatioErrorKind {
...
+
+impl<T: Integer + Signed + Bounded + NumCast + Copy> Ratio<T> {
```
cuviper: Can we reduce these bounds at all?  e.g. `Copy`, we usually only require `Clone`.

rational/src/lib.rs
```diff
@@ -774,6 +932,15 @@ mod test {
...
     }
 
+    #[test]
+    fn test_approximate_float() {
+        assert_eq!(super::approximate_float(0.5f32, 1.0e-20, 30), Some(Ratio::new(1i64, 2)));
+        assert_eq!(super::approximate_float(0.5f64, 1.0e-20, 30), Some(Ratio::new(1i32, 2)));
+        assert_eq!(super::approximate_float(5f32, 1.0e-20, 30), Some(Ratio::new(5i64, 1)));
+        assert_eq!(super::approximate_float(5f64, 1.0e-20, 30), Some(Ratio::new(5i32, 1)));
+        assert_eq!(super::approximate_float(29.97f32, 1.0e-20, 30), Some(Ratio::new(2997i64, 100)));
+        assert_eq!(super::approximate_float(-29.97f32, 1.0e-20, 30), Some(Ratio::new(-2997i64, 100)));
```
cuviper: Need some `None` tests in here.

rational/src/lib.rs
sdroege: Clone instead of Copy would be possible, yes. The others are all needed unless you can think of another way for the conversions between T and F and back

rational/src/lib.rs
sdroege: Good point! I'll update this later, thanks for the review

rational/src/lib.rs
sdroege: Adding u8 to u64, and usize/isize is not a problem. Will do so later.

However what would be missing then is a generic way of converting of unsigned integers due to the trait bound overlaps we discussed before.

rational/src/lib.rs
sdroege: Let me know if you want more tests. Might make sense to add something around quickcheck here at some point.

rational/src/lib.rs
```diff
@@ -668,6 +668,176 @@ impl RatioErrorKind {
...
+// No Unsigned constraint because this also works on positive integers and is called
+// like that, see above
+fn approximate_float_unsigned<T, F>(val: F, max_error: F, max_iterations: usize) -> Option<Ratio<T>>
+    where T: Integer + Bounded + NumCast + Clone,
+          F: Float + NumCast
+{
+    // Continued fractions algorithm
+    // http://mathforum.org/dr.math/faq/faq.fractions.html#decfrac
+
+    assert!(val >= F::zero());
```
cuviper: I think the unsigned `FromPrimitive` need to deal with negative values before they reach this assert!

rational/src/lib.rs
cuviper: How about some floating point boundaries?  NaN, infinity, and even the finite yet absurdly large and small.  And unsigned ratios need testing now too, including that behavior with negative floats.

I'll bet we could benefit from quickcheck throughout num!

rational/src/lib.rs
cuviper: Yeah, that limitation is OK for now.

rational/src/lib.rs
cuviper: Well, maybe using explicit `ToPrimitive` + `FromPrimitive` is slightly better.  Not everything that implements those will also implement `NumCast`, unfortunately.

It's a weird trait, e.g. that `NumCast` requires `Self: ToPrimitive`, and not `FromPrimitive`, but otherwise it seems like something that really ought to have a blanket `impl`.  (breaking changes...)

rational/src/lib.rs
sdroege: Not sure I understand what you mean here. The signed ones all go into approximate_float(), which calls approximate_float_unsigned() for the absolute value and then negates the result if the original value was negative. approximate_float_unsigned() asserts that the input value is >= 0, i.e. "unsigned".

cuviper: For `<u64 as FromPrimitive>::from_f64(-1.0)`, I expect `None`.  Is that the case?

sdroege: Ah I see now what you mean, thanks!

rational/src/lib.rs
sdroege: Done, quickcheck would be for another time

rational/src/lib.rs
sdroege: > For <u64 as FromPrimitive>::from_f64(-1.0), I expect None. Is that the case?

It is now, yes

sdroege:
Are there any other changes you'd like to see?

cuviper:
Looks really nice, thanks!

@homu r+

PR #287, opened 2017-04-30
Replace deprecated Hasher

patricksjackson:
`std::hash::SipHasher` is now deprecated, see https://doc.rust-lang.org/std/hash/struct.SipHasher.html


cuviper:
This would be fine, but as you can see from CI we'd have to increase our minimum Rust.  `DefaultHasher` wasn't stable until 1.13.

Since this is just for testing, maybe instead we can hash them indirectly?  e.g. just try to use these types with a default `HashSet`, without having to specify the exact hasher being used.

PR #291, opened 2017-05-07
Get the default hasher indirectly

cuviper:
`DefaultHasher` wasn't stable until 1.13, at which point all the other
hashers were deprecated, so it's not easy for us to name a hasher type to
use for testing.  However, `RandomState` has been stable since 1.7, and it
implements `BuildHasher` that has a `Hasher` associated type.

(extends #287)


cuviper:
It's messy, but it's just for testing anyway... :)

@homu r+

PR #292, opened 2017-05-15
Implement associated trait const for ZERO / ONE where possible, behind default-off feature gate

chrivers:
For me, the biggest missing feature in num is the ability to use Zero() / One() without incurring overhead. The inline hints probably cause many uses to optimize away, but this isn't true in Debug builds, at least.

Without better support from the compiler (for consts with destructors, specifically), it will not be possible to add a blanket impl for the const version. For that reason, I have made 2 new traits, ZeroConst and OneConst, which are feature-gated, and off by default.

For performance-critical applications, these can be added as trait bounds, and will provide ::ZERO and ::ONE as Self-type consts. Here's a snippet from an actual program, using these new traits:

```rust
use num::{Integer, ZeroConst, OneConst};

// ....

impl<T> Repr<T> for bool where
    T: Integer + ZeroConst + OneConst
{
    fn decode(x: T) -> Self { x > T::ZERO }
    fn encode(self) -> T { if self { T::ONE } else { T::ZERO } }
}
```

This probably should never escape the feature gate, but it provides a very valuable way for users needing the const version to get it, without resorting to worse hackery. I look forward to hearing your feedback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,8 @@ complex = ["num-complex"]
 rational = ["num-rational"]
 default = ["bigint", "complex", "rational", "rustc-serialize"]
 
+experimental_trait_consts = []
+
 serde = [
   "num-bigint/serde",
   "num-complex/serde",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,8 @@ pub use num_rational::BigRational;
 pub use num_complex::Complex;
 pub use num_integer::Integer;
 pub use num_iter::{range, range_inclusive, range_step, range_step_inclusive};
+#[cfg(feature="experimental_trait_consts")]
+pub use num_traits::{ZeroConst, OneConst};
 pub use num_traits::{Num, Zero, One, Signed, Unsigned, Bounded,
                      one, zero, abs, abs_sub, signum,
                      Saturating, CheckedAdd, CheckedSub, CheckedMul, CheckedDiv,
--- a/traits/Cargo.toml
+++ b/traits/Cargo.toml
@@ -10,3 +10,6 @@ name = "num-traits"
 version = "0.1.37"
 
 [dependencies]
+
+[features]
+"experimental_trait_consts" = []
--- a/traits/src/identities.rs
+++ b/traits/src/identities.rs
@@ -25,6 +25,11 @@ pub trait Zero: Sized + Add<Self, Output = Self> {
     fn is_zero(&self) -> bool;
 }
 
+#[cfg(feature="experimental_trait_consts")]
+pub trait ZeroConst: Zero {
+    const ZERO: Self;
+}
+
 macro_rules! zero_impl {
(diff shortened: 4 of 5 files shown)


cuviper:
There may be a niche case to have this in a crate somewhere, but I'm not inclined for `num`.

- I don't want to get into nightly features, namely `#[feature(associated_consts)]`.  The only other places where we use nightly features are for benchmarking (`#[feature(test)]`) and the deprecated `num-macros`.
- Our Rust backward-compatibility is currently at 1.8, and Travis CI shows that this addition doesn't compile at all, even though it's behind a disabled `cfg(feature)`.  I'd call that a compiler bug, but still one we have to live with.
- I think it's a mistake to worry about the overhead of `zero()` and `one()` in debug builds *at all*.  There's so much that will cause your program to run slowly that I doubt these are even noticeable, unless you craft a microbenchmark for just this case.
- In release builds, I'd be really surprised to see any performance difference between the functions and the constants.  They really should be inlined away almost every time, so much that I'd consider filing a rustc bug if they ever showed up in a performance profile.

chrivers:
Thanks for the feedback.

As you say, it doesn't make sense that a conditional compilation that is disabled, would fail on 1.8, but I guess that's the reality we have to live with. I tested with stable and nightly, and it worked fine there. As such, I can't support merging this PR, obviously.

However, I'd just like to point out that it does make a real difference. I'm not sure it does in release mode, but for the projects I've worked on so far, it's absolutely infeasible to do compile cycles in release mode, because of the compile times.

In my current project, I run a blackbox test across a varied range of packets that needs parsing. Changing between fns and consts makes about a 10% performance difference in debug builds - for the *whole* program. It's certainly enough to be annoying.

Ok, so we can't merge this as-is. I still think the benefit is both real and useful. I wonder if 1.8 could stop worrying about the syntax if I put all the const code in a separate module, so the whole module could be feature gated?

Are you opposed to the idea, if we could find a clean solution for supporting 1.8?

cuviper:
I also don't want to get into nightly features, nor add more configuration features of our own.  It's already annoying to manage the combinations of features we have now.

You could easily publish this in your own tiny crate, or even just embedded inside your own code, since as I said this feels rather niche to me to worry about debug performance.  I believe you that it makes a difference in your testing, but I think that's not compelling enough to add maintenance here in `num`.

chrivers:
Fair enough, that seems like a good solution. If I publish such a crate, would you be willing to include maybe a line of text and link in the readme (or whereever it makes sense), so people know where to go if they need it?

cuviper:
Sure, we should gather some sort of "related crates" section.

davll:
Associated constants feature has already landed on stable rustc now.

shepmaster:
A question was asked a while ago on Stack Overflow; here's [an example of a trait that can add associated constants](https://stackoverflow.com/a/32763878/155423) on top of e.g. the `Float` trait. This does work in stable Rust nowadays.

PR #293, opened 2017-05-18
Adds quickcheck::Arbitrary impls for BigInt, BigUint, Complex, and Ratio.

### LINKED ISSUE #241: Quickcheck Arbitrary support?
Would there be interest in supporting (perhaps via a feature) the Quickcheck Arbitrary trait?

An example:

```rust
impl Arbitrary for BigUint {
  fn arbitrary<G: Gen>(g: &mut G) -> Self {
    BigUnit::new(g.gen::<Vec<u32>>())
}
```

Reasoning:

* As a user I would like to be able to generate arbitrary values for use in `quickcheck! {}` based tests.
* It is not possible to implement `Arbitrary` on `BigUint`, for example, since it it part of a different crate.

I would be interested in working on this PR but I want to make sure it would be welcome.

remexre:
Resolves #241. Gated by the `quickcheck` feature.

Would it be okay to do a release after this is merged?


cuviper:
Sorry I haven't reviewed this yet, but I just want you to know I'm not ignoring it.  I need to familiarize myself with how quickcheck actually works. :)

One request up front: can you add some tests that exercise the new code?  First just some basic tests to make sure the new trait implementations actually work at all.  Then I imagine it should also be possible to write some useful quickcheck tests to validate num itself.

remexre:
Yep, I can do that. I'm actually working on landing `#[derive(Arbitrary)]` in the quickcheck crate (https://github.com/BurntSushi/quickcheck/issues/98); I'll probably might make this use that instead.

I can definitely add the tests, although my schedule's a bit busy until the weekend.

cuviper:
Did you ever get anywhere with this?  If you'd like to continue, please move them to the separate repos.

The new separate issues are:
- rust-num/num-bigint#5
- rust-num/num-complex#1
- rust-num/num-rational#2

remexre:
I haven't gotten around to getting BurntSushi/quickcheck#175 merged yet; I'll get around to it at some point...

Issue #216, opened 2016-08-09
Make subcrates `#![no_std]`


hauleth:
- [ ] `num-traits`
- [ ] `num-integer`
- [ ] `num-complex`
- [ ] `num-iter`

cuviper:
I noticed in the 1.11 release notes that `to_degrees` and `to_radians` were just now added to libcore.  Things like this will constrain our rust version for going `no_std`.  (But when we do make the breaking transition, we'll be choosing a new rust base anyway.)

cmr:
Any timeline on when this will be done? Especially num-traits?

Kixunil:
I'm interested in this too.

`to_radians` and `to_degrees` can be implemented easily without Rust support. It's simply `self * PI / 180.0` and `self * 180.0 / PI` (with appropriate conversion).

vks:
I started working on porting `num-traits` to `no_std`. However, I run into a very irritating problem:
```
error[E0308]: mismatched types
   --> src/sign.rs:116:38
    |
116 |                 <$t>::abs_sub(*self, *other)
    |                                      ^^^^^^ expected &f64, found f64
...
141 | signed_float_impl!(f64, f64::NAN, f64::INFINITY, f64::NEG_INFINITY);
    | -------------------------------------------------------------------- in this macro invocation
    |
    = note: expected type `&f64`
               found type `f64`
    = help: try with `&*other`
```
Removing the `*` does not work:
```
error[E0308]: mismatched types
   --> src/sign.rs:116:38
    |
116 |                 <$t>::abs_sub(*self, other)
    |                                      ^^^^^ expected f64, found &f64
...
141 | signed_float_impl!(f64, f64::NAN, f64::INFINITY, f64::NEG_INFINITY);
    | -------------------------------------------------------------------- in this macro invocation
    |
    = note: expected type `f64`
               found type `&f64`
```
So the compiler expects `f64` and `&f64` in the same place?

vks:
Even fixing that problem by reimplementing `abs_sub` it still does not work:
```
error: no method named `is_finite` found for type `f64` in the current scope
   --> src/cast.rs:231:19
    |
231 |             if !n.is_finite() || (-max_value as f64 <= n && n <= max_value as f64) {
    |                   ^^^^^^^^^
...
273 | impl_to_primitive_float!(f32);
    | ------------------------------ in this macro invocation
    |
[...]

PR #296, opened 2017-05-31
traits: Introduce std feature

vks:
This makes it possible to build `traits` without `std`. For this a new
trait `BasicFloat` was introduced, implementing some basic functionality
that works with `core`. Most notably this is lacking functions like
`cos`, `sin`, etc.

`Float` is not available without `std`.

Refs #216.

### DIFF
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -7,199 +7,6 @@ use core::f32;
 
 use {Num, NumCast};
 
-/// Basic floating point operations that work with `core`.
-pub trait BasicFloat: Num + Neg<Output = Self> + PartialOrd + Copy {
-    /// Returns positive infinity.
-    #[inline]
-    fn infinity() -> Self;
-
-    /// Returns negative infinity.
-    #[inline]
-    fn neg_infinity() -> Self;
-
-    /// Returns NaN.
-    #[inline]
-    fn nan() -> Self;
-
-    /// Returns `true` if the number is NaN.
-    #[inline]
-    fn is_nan(self) -> bool {
-        self != self
-    }
-
-    /// Returns `true` if the number is infinite.
-    #[inline]
-    fn is_infinite(self) -> bool {
-        self == Self::infinity() || self == Self::neg_infinity()
-    }
-
-    /// Returns `true` if the number is neither infinite or NaN.
-    #[inline]
-    fn is_finite(self) -> bool {
-        !(self.is_nan() || self.is_infinite())
-    }
-
-    /// Returns `true` if the number is neither zero, infinite, subnormal or NaN.
-    #[inline]
-    fn is_normal(self) -> bool {
-        self.classify() == FpCategory::Normal
-    }
-
-    /// Returns the floating point category of the number. If only one property
-    /// is going to be tested, it is generally faster to use the specific
-    /// predicate instead.
-    #[inline]
-    fn classify(self) -> FpCategory;
-
-    /// Computes the absolute value of `self`. Returns `BasicFloat::nan()` if the
-    /// number is `BasicFloat::nan()`.
-    #[inline]
-    fn abs(self) -> Self {
-        if self.is_sign_positive() {
-            return self;
-        }
-        if self.is_sign_negative() {
-            return -self;
-        }
-        Self::nan()
-    }
-
-    /// Returns a number that represents the sign of `self`.
-    ///
-    /// - `1.0` if the number is positive, `+0.0` or `BasicFloat::infinity()`
-    /// - `-1.0` if the number is negative, `-0.0` or `BasicFloat::neg_infinity()`
-    /// - `BasicFloat::nan()` if the number is `BasicFloat::nan()`
-    #[inline]
-    fn signum(self) -> Self {
-        if self.is_sign_positive() {
-            return Self::one();
-        }
-        if self.is_sign_negative() {
-            return -Self::one();
-        }
-        Self::nan()
-    }
-
-    /// Returns `true` if `self` is positive, including `+0.0` and
-    /// `BasicFloat::infinity()`.
-    #[inline]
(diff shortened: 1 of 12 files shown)


cuviper:
There's a CI failure reported across all versions.  I think this probably needs some direct no-std testing in `ci/test_full.sh` too.

Did you consider just gating `#[cfg(feature = "std")]` on `Float` methods as needed?  I know that makes a lot more places that have to be marked, but I think it's easier to explain.  I don't like the method duplication, and I don't want users to be confused about when to use `Float` or `BasicFloat`.

If you really want to have two separate traits, maybe it would be better to have `Float: BasicFloat` so they are directly related.

cuviper:
> maybe it would be better to have `Float: BasicFloat`

Hmm, but that's a breaking change for implementors. :/

vks:
> Hmm, but that's a breaking change for implementors. :/

Yes, I considered that, but discarded it for this reason.

I like your suggestion of gating the methods, this would have the advantage that there will be no redundant reimplementation of the methods currently in `BasicFloat` if `std` is available.

vks:
I did what you suggested, which coincidentally also fixed the tests.

I added quite a few default implementations to `Float`. They only matter for `no_std` at the moment.

traits/src/float.rs
```diff
@@ -402,7 +455,34 @@ pub trait Float
...
+        let mut acc = base;
+        while exp > 1 {
+            exp >>= 1;
+            base = base * base;
+            if exp & 1 == 1 {
+                acc = acc * base;
+            }
+        }
+        acc
+    }
```
cuviper: Maybe use the generic `num_traits::pow()` once `exp` is positive?  I see your implementation is identical, but we don't need to duplicate code.

traits/src/float.rs
```diff
@@ -537,13 +625,28 @@ pub trait Float
...
+    /// use std::f64::consts;
+    ///
+    /// let angle = consts::PI;
+    ///
+    /// let abs_difference = (angle.to_degrees() - 180.0).abs();
+    ///
+    /// assert!(abs_difference < 1e-10);
+    /// ```
+    #[cfg(not(feature = "std"))]
+    fn to_degrees(self) -> Self;
```
cuviper: Ugh, that's annoying to have a separate declaration, but I don't see a better way, other than excluding it from `no_std` entirely.

traits/src/float.rs
```diff
@@ -1097,117 +1287,138 @@ macro_rules! float_impl {
             fn to_degrees(self) -> Self {
                 // NB: `f32` didn't stabilize this until 1.7
                 // <$T>::to_degrees(self)
-                self * (180. / ::std::$T::consts::PI)
+                self * (180. / ::core::$T::consts::PI)
```
cuviper: Our baseline is now 1.8 - can we just use the native implementation?

traits/src/float.rs
```diff
@@ -1097,117 +1287,138 @@ macro_rules! float_impl {
...
             }
 
             #[inline]
             fn to_radians(self) -> Self {
                 // NB: `f32` didn't stabilize this until 1.7
                 // <$T>::to_radians(self)
-                self * (::std::$T::consts::PI / 180.)
+                self * (::core::$T::consts::PI / 180.)
```
cuviper: Ditto on using the native impl.

traits/src/float.rs
vks: The problem is that `num_traits::pow` takes a `usize` for the exponent, which might be smaller than `i32` in the future. That is why I duplicated the implementation.

If we made a breaking change, then we could consolidate the two implementations by making `num_traits::pow` use a `u32` for the exponent.

cuviper: Hmm, do 16-bit targets even have these floats?  But OK, not a big deal.

vks: Probably not. I could use `num_traits::pow` and cast, hoping it works. Alternatively we could use a macro to deduplicate the implementation.

cuviper: Or instead of casting, use `FromPrimitive` and `unwrap`.

vks:
@cuviper I think I addressed all your comments.

traits/src/float.rs
```diff
@@ -950,264 +1094,312 @@ macro_rules! float_impl {
...
             fn log10(self) -> Self {
                 <$T>::log10(self)
             }
 
             #[inline]
             fn to_degrees(self) -> Self {
-                // NB: `f32` didn't stabilize this until 1.7
-                // <$T>::to_degrees(self)
-                self * (180. / ::std::$T::consts::PI)
+                <$T>::to_degrees(self)
```
cuviper: Oh, this doesn't work for no_std, because the inherent methods are in std.  So it recurses:

```rust
warning: function cannot return without recurring
    --> src\float.rs:1272:13
     |
1272 |               fn to_degrees(self) -> Self {
     |  _____________^ starting here...
1273 | |                 <$T>::to_degrees(self)
1274 | |             }
     | |_____________^ ...ending here
...
1482 |   float_impl!(f32 integer_decode_f32 classify_f32);
     |   ------------------------------------------------- in this macro invocation
     |
     = note: #[warn(unconditional_recursion)] on by default
```

I guess we have to roll our own, but it's not a big deal since these are trivial.

Can you add `#![deny(unconditional_recursion)]` to the crate?  This seems like a problem we're vulnerable to in general, with so many traits that are supposed to forward calls...

vks: Sorry, I actually wanted to keep the implementation for `no_std` but forgot about it...

Should be fixed in the most recent commit.

cuviper:
@homu r+

nox:
This is a breaking change and the release should be yanked. Feature-gating things that were previously not feature-gated is a breaking change.

PR #299, opened 2017-06-09
Fix building without the num-bigint feature

dotdash:


cuviper:
Yeah, that's correct.  I'm working on the change with broader CI testing, but we can merge this first.  Thanks!

@homu r+

PR #300, opened 2017-06-09
Test and fix subcrates with --no-default-features

cuviper:
- The CI script now tests them all with --no-default-features.
- bigint: testing needs rand even when the main feature is disabled.
- rational: gate `FromPrimitive for Ratio<BigInt>` on having bigint.


cuviper:
@homu r+

Issue #281, opened 2017-04-23
Update to Serde 1.0


dtolnay:
Release notes for 0.9.0: https://github.com/serde-rs/serde/releases/tag/v0.9.0
Release notes for 1.0.0: https://github.com/serde-rs/serde/releases/tag/v1.0.0

cuviper:
I think this is *likely* a breaking change, but if there's any way to conditionalize our code to cover a wide range of serde versions, maybe we can squeeze by.

dtolnay:
Technically it's definitely possible but I'm not going to recommend that you do it. :rofl:

The way to do it would be make a `serde-0-9` crate and a `serde-1-0` crate that each depend on the respective version of serde and just contain `extern crate serde; pub use serde::*`. Then you could provide an impl for `serde_0_9::Serialize` as well as `serde_1_0::Serialize`.

cuviper:
Oy.  I was thinking more along the lines of detecting the selected serde version in a build script somehow, and just emit a feature name for that.  But maybe that's not possible.

We need to do a semver bump.  I've been procrastinating that forever.  Maybe [stable i128](https://github.com/rust-lang/rust/issues/35118) will finally be worth advancing `num` so we don't have that major lack.

cuviper:
We could do it as a separate `num-serde` crate, like [`url`](https://crates.io/crates/url) does with [`url_serde`](https://crates.io/crates/url_serde).

corngood:
So I've got local changes for serde 1.0, and I was hoping to make a PR.

It seems like the options are: 

1) bump to 0.2 and require serde 1
2) support both with features (trying to wrap my head around that one)
3) support both with intermediate crates
4) wrapper crate (num_serde)

The changes are pretty trivial if we do (1), but I don't see a PR/branch with them.  Would you like me to push it so it can be linked to the 0.2 milestone?  If not, would you like me to take a look at one of the other options?

cuviper:
We need to just do 0.2.  I hate to force an upgrade on the ecosystem, but it's going to happen sometime.  We'll do it in a side branch so we can batch as many breaking changes as we can at once.

To that end, there's now a `next` branch, and #302 to get it going.  When that lands, you can send a PR for `next` instead of `master` with the breaking change.  Serde 1 only, still behind an optional feature, and probably remove rustc-serialize while we're at it.  Sound good?

cuviper:
This is done on the next branch.

PR #302, opened 2017-06-09
[next] Start development on 0.2

cuviper:
- Drop num-macros.
- Add a workspace.
- Bump all crates to 0.2.0-git, publish=false.
- Merge the history of master without losing num-traits `no_std`.


vks:
No that we can make breaking changes, we should probably reconsider how the `no_std` support was implemented for `Float`. We could for example make `Float` a subtrait of `NoStdFloat`.

cuviper:
Yeah, that would probably better to avoid the hazard about someone implementing only the core parts.  I suggest calling it `CoreFloat` or `FloatCore`.  It doesn't need to be in *this* PR though, as I intend to leave the branch open for a while before we publish.

vks:
Yes, there is quite a few other things we should implement in a breaking release, the generic algorithms should use the new `Num*` traits for instance.

rubdos:
Someone already took on #281 (Serde 1.0) ? Otherwise I would wet my hands on that one.

PR #303, opened 2017-06-11
Radix > 36 in BigUint

Phaiax:
Hey.

I have added two functions 

    pub from_radix(buf: &[u8], radix: u32) -> Option<BigUint>
    pub fn to_radix(&self, radix: u32) -> Vec<u8>

for conversion from and into radices > 36. Each input/output byte is one digit of the number in the base system.

I added tests and documentation.

Open Questions:
* Naming of the functions suboptimal? The human readable equivalents are named `from_str_radix()` / `parse_bytes()` and `to_str_radix()`.
* Order of input/output bytes. I know the arabic number system goes form the MSD to the LSD, but from a neutral view this is arbitrary and I chose the reverse order for from_radix() and to_radix(). This allows to summarize and multiplying by digit value without knowing the length of the digit. `base198bytes.iter().enumerate().fold(0, |sum, i, d| sum += d*radix^i)`.

Thanks,
Daniel


cuviper:
Do you really have need of larger radixes?  I suspect the only typical uses will be the "normal" 2..36 and a full-byte 256.  But I guess it's not too bad to go ahead and support more.

For the radix=256 case we do already have `to_` and `from_bytes_be` and `_le`.  So let's make the endianness an explicit choice in your functions too: `from_radix_be`, `from_radix_le`, `to_radix_be`, and `to_radix_le`.

Phaiax:
In particular I have the need for radix of 62. In the ascii-armoring scheme of [saltpack](https://saltpack.org/armoring), 32 bytes of raw data are interpreted as number and converted to 43 human readable characters.

I renamed the functions accordingly and reordered the functions, so that all new() like functions come first. Especially parse_bytes() is now more up front and gained a hint in the doc comment to the `fn from_str_radix` from the Num trait implementation.

clarcharr:
@cuviper supporting base64 might be desired in some cases!

Phaiax:
But base64 is very easy to implement even without the num crate because there is no bignum math involved. 4 base64 characters represent 3 bytes.

Phaiax:
With an extra argument for the sign? I thought BigInt isn't a drop in replacement for BigUint anyway because of the many different function signatures and free radix numbers do not have a sign component convention. I can do if you think the symmetry looks better, (But that ll take until Saturday,)

bigint/src/bigint.rs
```diff
@@ -980,6 +980,52 @@ impl BigInt {
...
+    ///
+    /// let inbase190 = vec![15, 33, 125, 12, 14];
+    /// let a = BigInt::from_radix_be(Sign::Minus, &inbase190, 190).unwrap();
+    /// assert_eq!(a.to_radix_be(190), (Sign:: Minus, inbase190));
+    /// ```
+    pub fn from_radix_be(sign: Sign, buf: &[u8], radix: u32) -> Option<BigInt> {
+        Some(BigInt::from_biguint(sign, match BigUint::from_radix_be(buf, radix) {
+            Some(u) => u,
+            None => { return None; }
+        }))
```
cuviper: This style seems rather convoluted -- why not have the `match` as the outer expression?  Or even just `map` the biguint result into a bigint.

Phaiax:
Rebased.

cuviper:
Very nice, thanks!

@homu r+

cuviper:
Hmm, has @homu finally died?  I may have to figure out bors-ng...

PR #304, opened 2017-06-16
Add BigInt two's complement conversions

Enet4:
This PR provides the methods `from_signed_bytes_le`, `from_signed_bytes_be`, `to_signed_bytes_le`, and `to_signed_bytes_be` to `BigInt`. This concern came around with [this SO question](https://stackoverflow.com/q/44345786/1233251), in which we realized that there was no readily available conversion from arrays in two's complement binary representation to a `BigInt`.
Please let me know what further changes should be made before it's good to go. In particular:

- Should the helper functions `twos_complement_(le|be)` go elsewhere? The algorithms module seems to be tied to `biguint` right now.
- The `from_...` methods currently require an extra copy by taking ownership of a `Vec<u8>`. Should this be avoided before it's merged?

### DIFF
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -1,5 +1,5 @@
 use std::default::Default;
-use std::ops::{Add, Div, Mul, Neg, Rem, Shl, Shr, Sub, Not, DerefMut};
+use std::ops::{Add, Div, Mul, Neg, Rem, Shl, Shr, Sub, Not};
 use std::str::{self, FromStr};
 use std::fmt;
 use std::cmp::Ordering::{self, Less, Greater, Equal};
@@ -16,7 +16,7 @@ use serde;
 use rand::Rng;
 
 use integer::Integer;
-use traits::{ToPrimitive, FromPrimitive, Num, WrappingAdd, CheckedAdd, CheckedSub,
+use traits::{ToPrimitive, FromPrimitive, Num, CheckedAdd, CheckedSub,
              CheckedMul, CheckedDiv, Signed, Zero, One};
 
 use self::Sign::{Minus, NoSign, Plus};
@@ -1214,16 +1214,14 @@ fn twos_complement_be(digits: &mut [u8]) {
 /// Perform in-place two's complement of the given digit iterator
 /// starting from the least significant byte.
 #[inline]
-fn twos_complement<T, R, I>(digits: I)
-    where I: IntoIterator<Item = R>,
-          R: DerefMut<Target = T>,
-          T: Clone + Not<Output = T> + WrappingAdd + Zero + One
+fn twos_complement<'a, I>(digits: I)
+    where I: IntoIterator<Item = &'a mut u8>
 {
     let mut carry = true;
     for mut d in digits {
-        *d = d.clone().not();
+        *d = d.not();
         if carry {
-            *d = d.wrapping_add(&T::one());
+            *d = d.wrapping_add(1);
             carry = d.is_zero();
         }
     }
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -1,5 +1,5 @@
 use std::default::Default;
-use std::ops::{Add, Div, Mul, Neg, Rem, Shl, Shr, Sub};
+use std::ops::{Add, Div, Mul, Neg, Rem, Shl, Shr, Sub, Not};
 use std::str::{self, FromStr};
 use std::fmt;
 use std::cmp::Ordering::{self, Less, Greater, Equal};
@@ -16,8 +16,8 @@ use serde;
 use rand::Rng;
 
 use integer::Integer;
-use traits::{ToPrimitive, FromPrimitive, Num, CheckedAdd, CheckedSub, CheckedMul,
-             CheckedDiv, Signed, Zero, One};
+use traits::{ToPrimitive, FromPrimitive, Num, WrappingAdd, CheckedAdd, CheckedSub,
+             CheckedMul, CheckedDiv, Signed, Zero, One};
 
 use self::Sign::{Minus, NoSign, Plus};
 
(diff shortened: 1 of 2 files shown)


bigint/src/bigint.rs
```diff
@@ -926,6 +926,56 @@ impl BigInt {
         BigInt::from_biguint(sign, BigUint::new(digits))
     }
 
+    /// Creates and initializes a `BigInt` from an array of bytes in two's complement.
+    ///
+    /// The digits are in little-endian base 2^8.
+    #[inline]
+    pub fn from_twos_complement_bytes_le(mut digits: Vec<u8>) -> BigInt {
```
cuviper: I do think this should take `&[u8]`, even if that means making extra copies for now.  We can work on optimizing that later, but the public interface needs to be stable.

bigint/src/bigint.rs
```diff
@@ -926,6 +926,56 @@ impl BigInt {
...
+        } else {
+            digits.iter().rev().next()
+                .map(|v| if *v > 127 {
+                    Sign::Minus
+                } else {
+                    Sign::Plus
+                })
+                // we have at least one non-zero digit
+                .unwrap()
+        };
```
cuviper: I'd just `match digits.last()` here, with `None => BigInt::zero()`, and let `from_biguint` take care of normalizing longer zero sequences.

bigint/src/bigint.rs
```diff
@@ -926,6 +926,56 @@ impl BigInt {
...
+        } else {
+            digits.iter().next()
+                .map(|v| if *v > 127 {
+                    Sign::Minus
+                } else {
+                    Sign::Plus
+                })
+                // we have at least one non-zero digit
+                .unwrap()
+        };
```
cuviper: Same idea, take `&[u8]` and `match digits.first()`.

bigint/src/bigint.rs
```diff
@@ -994,6 +1044,23 @@ impl BigInt {
...
+    /// ```
+    /// use num_bigint::ToBigInt;
+    ///
+    /// let i = -1125.to_bigint().unwrap();
+    /// assert_eq!(i.to_twos_complement_bytes_le(), vec![155, 251]);
+    /// ```
+    #[inline]
+    pub fn to_twos_complement_bytes_le(&self) -> Vec<u8> {
+        let mut bytes = self.data.to_bytes_le();
+        twos_complement_le(&mut bytes);
```
cuviper: You should only flip it for `Sign::Minus`!

It will also need to consider overflow - e.g. -192 will give you biguint `[0xc0]`, complemented should be `[0x40,0xff]`.

bigint/src/bigint.rs
```diff
@@ -1010,6 +1077,23 @@ impl BigInt {
...
+    /// ```
+    /// use num_bigint::ToBigInt;
+    ///
+    /// let i = -1125.to_bigint().unwrap();
+    /// assert_eq!(i.to_twos_complement_bytes_be(), vec![251, 155]);
+    /// ```
+    #[inline]
+    pub fn to_twos_complement_bytes_be(&self) -> Vec<u8> {
+        let mut bytes = self.data.to_bytes_be();
+        twos_complement_be(&mut bytes);
```
cuviper: Ditto on `Minus`, and -192 overflows to `[0xff, 0x40]`.

bigint/src/bigint.rs
```diff
@@ -1105,3 +1189,36 @@ impl BigInt {
         return Some(self.div(v));
     }
 }
+
+/// Perform in-place two's complement of the given digit range,
+/// in little-endian byte order.
+#[inline]
+fn twos_complement_le<T>(digits: &mut [T])
```
cuviper: I think it's harmful to clarity for this to be generic on `T`.  This is internal, we know it's `u8`.

But it might be beneficial to take `I: Iterator<Item=u8>` instead of a slice, and then we only need one copy of the function.  The `be` caller can `rev()` the iterator they pass in.

bigint/src/tests/bigint.rs
```diff
@@ -107,6 +107,43 @@ fn test_to_bytes_le() {
...
+    check(vec![0], "0");
+    check(vec![0; 10], "0");
+    check(vec![127, 255], "32767");
+    check(vec![255], "-1");
+    check(vec![1, 0, 0, 0], "16777216");
+    check(vec![156], "-100");
+    check(vec![128, 0, 0], "-8388608");
+    check(vec![255; 10], "-1");
+}
+
```
cuviper: Need some `to_` tests too...

bigint/src/bigint.rs
cuviper: Err, `Iterator<Item = &mut u8>` rather.

Enet4:
@cuviper Thanks for the review. I really did screw up in some places!

I think I managed to tackle each concern in the latest commit, but further feedback is still welcome.

bigint/src/bigint.rs
```diff
@@ -1105,3 +1196,35 @@ impl BigInt {
...
+    twos_complement(digits.iter_mut().rev())
+}
+
+/// Perform in-place two's complement of the given digit iterator
+/// starting from the least significant byte.
+#[inline]
+fn twos_complement<T, R, I>(digits: I)
+    where I: IntoIterator<Item = R>,
+          R: DerefMut<Target = T>,
+          T: Clone + Not<Output = T> + WrappingAdd + Zero + One
```
cuviper: I still don't see why this needs to be generic on `T`.  Can't we just use `u8` directly?  And `Item = &'a mut u8`?

Enet4: That has been addressed now. The aim of making it generic was to enable two's complement on arrays of `BigDigit`, thus moving this step into the internal representation and potentially avoiding the extra allocation from modifying the input byte array. I guess this can be done later with additional care.

cuviper: Hmm, well you didn't mention that intent when I brought it up before, or if you did I missed it.  Let's go with how you have it now, and later changes can freely reconsider since this isn't public.

cuviper:
Thanks!  I'll try @homu again here, but it didn't respond on #303, so I may have to merge+test manually...

@homu r+

PR #305, opened 2017-06-20
WIP: serde 1.0

rubdos:
W.r.t. https://github.com/rust-num/num/pull/302 & #218, I took the arrow to the knee and started changing what needed changing.

~~TODO:~~ Done; that was quicker than I thought.

- [x] BigInt
- [x] Complex
- [x] Rational

Did I miss more `serde` stuff?

### DIFF
--- a/bigint/Cargo.toml
+++ b/bigint/Cargo.toml
@@ -29,7 +29,7 @@ version = "0.3.19"
 
 [dependencies.serde]
 optional = true
-version = ">= 0.7.0, < 0.9.0"
+version = "1"
 
 [dev-dependencies.rand]
 version = "0.3.14"
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -70,7 +70,7 @@ impl Mul<Sign> for Sign {
 
 #[cfg(feature = "serde")]
 impl serde::Serialize for Sign {
-    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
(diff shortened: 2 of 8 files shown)


oli-obk:
Travis still has a `cargo update -p serde --precise 0.7.0` which probably needs to be updated ;)

rubdos:
> Travis still has a cargo update -p serde --precise 0.7.0 which probably needs to be updated ;)

Did I miss something else? :-)

oli-obk:
Rust 1.8 isn't supported anymore. Serde needs 0.13 I think (at least that's what our travis checks)

rubdos:
> Rust 1.8 isn't supported anymore. Serde needs 0.13 I think (at least that's what our travis checks)

So remove that too?

rubdos:
Seems like I had forgot a file.

rubdos:
Seem to pass now. Is there a reason though, to still support 1.13?

cuviper:
I consider it a breaking change to bump the minimum rust, so please don't
remove 1.8 testing entirely here. Just add 1.13 to CI as the minimum for
using serde.

(Even though we *are* making a breaking change here, let's isolate that
change to serde for the moment.)

bigint/Cargo.toml
```diff
@@ -29,7 +29,7 @@ version = "0.3.19"
 
 [dependencies.serde]
 optional = true
-version = ">= 0.7.0, < 0.9.0"
+version = ">= 1.0.0, < 2.0.0"
```
cuviper: Please make this a simple "1.0" dependency.

cuviper:
Thanks!

PR #311, opened 2017-06-29
rational: make sure Hash agrees with Eq

### LINKED ISSUE #310: num_rational::Ratio's Hash is incompatible with Eq.
As of 0.1.38, `num_rational::Ratio` implements `PartialEq` and `Eq` by the numeric value, while `Hash` is derived. This violates [the rule of Eq + Hash](https://doc.rust-lang.org/core/hash/trait.Hash.html#hash-and-eq) which is `a == b ⇒ hash(a) == hash(b)` when non-reduced ratio is involved:

```rust
let a = Rational::new_raw(4, 2);
let b = Rational::new_raw(6, 3);
assert_eq!(a, b); // ok.

let mut h = HashSet::new();
h.insert(a);
h.insert(b);
assert_eq!(h.len(), 1); // likely panic, 2 ≠ 1
```

The hash should be computed using the reduced value. This unfortunately means the GCD will need to be calculated every time we need the hash.

cuviper:
We can't use a derived `Hash` when we have a manual `Eq`, because we
need to uphold the invariant `a == b` → `h(a) == h(b)`.  Since `Eq`
doesn't require them to be in reduced form, `Hash` also needs to be
normalized.

Fixes #310.

### DIFF
--- a/rational/src/lib.rs
+++ b/rational/src/lib.rs
@@ -27,8 +27,7 @@ extern crate num_integer as integer;
 use std::cmp;
 use std::error::Error;
 use std::fmt;
-#[cfg(test)]
-use std::hash;
+use std::hash::{Hash, Hasher};
 use std::ops::{Add, Div, Mul, Neg, Rem, Sub};
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


cuviper:
**Technically** this is a breaking change, because the former derived `Hash` only required `T: Hash`, and here I've used `T: Clone + Integer + Hash`.  However, every other operation on `Ratio<T>` requires `Clone + Integer`, and we can't fix the relationship to `Eq` with just `T: Hash`.

(I don't actually need `Clone` here, but I think it's better to be consistent.)

cuviper:
There's not even a way to create a `Ratio<T>` without `T: Clone + Integer` except through deserialization.  I think the only practical way this will break is for a generic function that requires only `T: Hash` but expects to use `Ratio<T> as Hash` from that.  I think this is outweighed by the need to fix `Hash` related to `Eq`.

Thanks to @mbrubeck for reviewing this with me on IRC.

bors r+

PR #313, opened 2017-06-29
Scalar operations across all integer types

feadoor:
With my apologies for opening a new PR, and also for the 8 month delay, this continues the work started in #237 - the discussion there outlines the goals I was aiming for. I suppose this supersedes that PR and the other one can now be closed.

This PR adds support for Add, Sub, Mul, Div and Rem operations involving one BigInt/BigUint and one primitive integer, with operands in either order, and any combination of owned/borrowed arguments.

### DIFF
--- a/bigint/src/biguint.rs
+++ b/bigint/src/biguint.rs
@@ -519,8 +519,7 @@ impl Div<BigUint> for BigDigit {
     type Output = BigUint;
 
     #[inline]
-    fn div(self, mut other: BigUint) -> BigUint {
-        other = other.normalize();
+    fn div(self, other: BigUint) -> BigUint {
         match other.data.len() {
             0 => panic!(),
             1 => From::from(self / other.data[0]),
@@ -557,8 +556,7 @@ impl Rem<BigUint> for BigDigit {
     type Output = BigUint;
 
     #[inline]
-    fn rem(self, mut other: BigUint) -> BigUint {
-        other = other.normalize();
+    fn rem(self, other: BigUint) -> BigUint {
         match other.data.len() {
             0 => panic!(),
             1 => From::from(self % other.data[0]),
--- a/bigint/src/biguint.rs
+++ b/bigint/src/biguint.rs
@@ -396,6 +396,7 @@ impl<'a> Add<&'a BigUint> for BigUint {
 
 promote_unsigned_scalars!(impl Add for BigUint, add);
 forward_all_scalar_binop_to_val_val_commutative!(impl Add<BigDigit> for BigUint, add);
+forward_all_scalar_binop_to_val_val_commutative!(impl Add<DoubleBigDigit> for BigUint, add);
 
 impl Add<BigDigit> for BigUint {
     type Output = BigUint;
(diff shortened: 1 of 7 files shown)


cuviper:
Great!  I've been meaning to ping you about this. :)

Before I dig into a full review, I'm having second thoughts about doing this for *all* integer types, because of this one usability wart: type inference fails for unspecified literals.

```
error[E0277]: the trait bound `num::BigUint: std::ops::Add<i32>` is not satisfied                          
 --> examples/bigint.rs:6:13                                                                               
  |                                                                                                        
6 |     let x = BigUint::zero() + 42;                                                                      
  |             ^^^^^^^^^^^^^^^^^^^^ the trait `std::ops::Add<i32>` is not implemented for `num::BigUint`  
  |                                                                                                        
  = note: no implementation for `num::BigUint + i32`                                                       
```

(that example file isn't committed here, just something I wrote to try it out.)

When rustc can't infer the exact type for an integer literal, it just chooses `i32`, even though that's not one of the actual possibilities for `BigUint`.  The options I see for us are:

- accept that limitation; users will always have to write a specific unsigned literal type, e.g. `42u32`.
- go ahead and implement `Add<i32> for BigUint`, and treat negatives like overflow
  - maybe mapping to subtraction where possible, and map `Sub<i32>` negatives to addition.
- step back and just implement `u32` for `BigUint`.
  - `BigInt` can have both `u32` and `i32`, as the inference fallback fits in the latter.

What do you think?

feadoor:
What a funny little error! Of your three options, I prefer the first. The compiler error is very clear about what's going on, and it doesn't seem unreasonable to be expected to add the type suffix in this case.

>  - go ahead and implement Add\<i32\> for BigUint, and treat negatives like overflow
>      - maybe mapping to subtraction where possible, and map Sub\<i32\> negatives to addition.

I don't like this option. Either `i32` becomes a special case, which is confusing to users, or all signed types get implemented for `BigUint`, which encourages bad behaviour from users - signed primitives should belong with `BigInt`, and not with `BigUint`.

> - step back and just implement u32 for BigUint.
>     - BigInt can have both u32 and i32, as the inference fallback fits in the latter.

I think it's more irritating / confusing for a user to think "Why can't I do this operation on a `u64`?" than it is for them to be told by the compiler that they need to specify the type of a literal.

bigint/src/biguint.rs
```diff
@@ -394,6 +396,48 @@ impl<'a> Add<&'a BigUint> for BigUint {
...
+impl Add<BigDigit> for BigUint {
+    type Output = BigUint;
+
+    #[inline]
+    fn add(mut self, other: BigDigit) -> BigUint {
+        if self.data.len() == 0 && other != 0 {
+            self.data.push(0);
+        }
+
+        let carry = __add2(&mut self.data, &[other]);
```
cuviper: For `len() == 0 && other == 0`, won't this fail the `__add2` length assertion?
(could just return `self` right away for `other == 0`, of course)

bigint/src/biguint.rs
```diff
@@ -394,6 +396,48 @@ impl<'a> Add<&'a BigUint> for BigUint {
...
+    fn add(mut self, other: DoubleBigDigit) -> BigUint {
+        if self.data.len() == 0 && other != 0 {
+            self.data.push(0);
+        }
+        if self.data.len() == 1 && other > BigDigit::max_value() as DoubleBigDigit {
+            self.data.push(0);
+        }
+
+        let (lo, hi) = big_digit::from_doublebigdigit(other);
+        let carry = __add2(&mut self.data, &[lo, hi]);
```
cuviper: Similarly here, the `__add2` length assertion doesn't care if they're zeros.

Issue #312, opened 2017-06-29
assertion failed: !nan.is_sign_positive() && !nan.is_sign_negative()


cuviper:
The standard library has changed behavior: issue rust-lang/rust#42425, pr rust-lang/rust#42431

This breaks our tests on nightly:

```text
---- src/float.rs - float::Float::is_sign_positive (line 329) stdout ----
	thread 'rustc' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: !nan.is_sign_positive() && !nan.is_sign_negative()', <anon>:14
note: Run with `RUST_BACKTRACE=1` for a backtrace.

', /checkout/src/librustdoc/test.rs:318
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- src/float.rs - float::Float::is_sign_negative (line 348) stdout ----
	thread 'rustc' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: !nan.is_sign_positive() && !nan.is_sign_negative()', <anon>:14
note: Run with `RUST_BACKTRACE=1` for a backtrace.

', /checkout/src/librustdoc/test.rs:318


failures:
    src/float.rs - float::Float::is_sign_negative (line 348)
    src/float.rs - float::Float::is_sign_positive (line 329)
```

I think we'll have to drop our own expectations about NaN here, and just let it follow whatever std does, whether old or new.  We also have our own implementations for `no_std` on the `next` branch, which should probably follow the new behavior.

PlasmaPower:
I'm working on this.

cuviper:
This is fixed on both master and next now.

PR #314, opened 2017-07-08
Derive ToPrimitive for enums

PlasmaPower:
I had to double the compile fail tests, as they will bail on the first error.

I have some ideas for more complex to/from primitive derives (with inner values that implement to/from primitive), but I'll save those for a future PR.


PlasmaPower:
Failure unrelated to this PR, it's instead due to https://github.com/rust-lang/rust/pull/42431

cuviper:
That failure is also #312.

Simple enums can be cast directly to integers with `as`.  I do see the value in implementing this to support generic programming with `ToPrimitive`, but it should probably just use `as` internally.

PlasmaPower:
@cuviper Oh I had no idea! I'll change the implementation.

PlasmaPower:
`as` requires a value, not just a reference. We have no guarantee that the enum implements clone. It seems the current implementation is the only way.

cuviper:
Hmm, that's awkward. Such enums should be trivially copyable.

cuviper:
> It's illegal even if they do implement Copy, due to macro hygiene.

I don't think it's quite *that* hygienic -- at least from a quick test, deriving with `as` on a `Copy` enum seems to work for me.  But still, we can't assume `Copy`.  Maybe someday we can specialize it.

bors r+

PlasmaPower:
@cuviper I didn't check that, but trying to use the `as` operator errors either way because the compiler doesn't know if the enum has associated values. Even though we check that, the compiler doesn't know we check that. It gives an error about the enum not being a primitive type, and suggests using `.into()`.

cuviper:
That's ok.  What you wrote *works*, most importantly.  Thanks!

PR #315, opened 2017-07-08
Fix float NaN positive/negative assumptions

PlasmaPower:
These are the minimal assumptions to make about `NaN`. Fixes part of #312 (but not the `next` branch).

### DIFF
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -330,14 +330,14 @@ pub trait Float
     /// use num_traits::Float;
     /// use std::f64;
     ///
-    /// let nan: f64 = f64::NAN;
+    /// let neg_nan: f64 = -f64::NAN;
     ///
     /// let f = 7.0;
     /// let g = -7.0;
     ///
     /// assert!(f.is_sign_positive());
     /// assert!(!g.is_sign_positive());
-    /// assert!(!nan.is_sign_negative());
+    /// assert!(!neg_nan.is_sign_positive());
     /// ```
     fn is_sign_positive(self) -> bool;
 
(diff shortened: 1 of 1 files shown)


traits/src/float.rs
```diff
@@ -323,8 +323,8 @@ pub trait Float
     /// ```
     fn signum(self) -> Self;
 
-    /// Returns `true` if `self` is positive, including `+0.0` and
-    /// `Float::infinity()`.
+    /// Returns `true` if `self` is positive, including `+0.0`,
+    /// `Float::infinity()`, and `f64::NAN` with newer versions of Rust.
```
PlasmaPower: Any ideas on how to make this less confusing? Right now it's not clear what "newer versions of Rust" applies to.

PlasmaPower: NVM ".., and with newer versions of Rust `f64::NAN`" works.

traits/src/float.rs
```diff
@@ -337,27 +337,25 @@ pub trait Float
     ///
     /// assert!(f.is_sign_positive());
     /// assert!(!g.is_sign_positive());
-    /// // Requires both tests to determine if is `NaN`
-    /// assert!(!nan.is_sign_positive() && !nan.is_sign_negative());
+    /// assert!(!nan.is_sign_negative());
```
cuviper: Hmm, we should be testing `fn is_sign_positive()` here, but I realize older rust won't return true.  Perhaps this one should check `!neg_nan.is_sign_positive()`.

traits/src/float.rs
```diff
@@ -337,27 +337,25 @@ pub trait Float
...
+    /// let neg_nan: f64 = -f64::NAN;
     ///
     /// let f = 7.0;
     /// let g = -7.0;
     ///
     /// assert!(!f.is_sign_negative());
     /// assert!(g.is_sign_negative());
-    /// // Requires both tests to determine if is `NaN`.
-    /// assert!(!nan.is_sign_positive() && !nan.is_sign_negative());
+    /// assert!(!neg_nan.is_sign_positive());
```
cuviper: And this doctest should have the `!nan.is_sign_negative()`.

cuviper:
Thanks!

I'm not sure if the bot will work since it's not on your branch yet, but let's try:

bors r+

PR #316, opened 2017-07-09
Enable bors-ng

cuviper:


cuviper:
Bah, #312.  But bors is alive, so let's merge anyway.

Issue #289, opened 2017-05-01
Implement FromStr for Complex


c410-f3r:
Would be nice to implement `FromStr` for `Complex`, is something blocking this feature?

```
let s = "3.0850107479750e-04 -4.9785737087053e-05";
let x = Complex::from_str(s).unwrap();
```

I could do some research and create a PR if you guys don't mind.

cuviper:
Off hand, I don't know a good reason why not!  There are probably a lot of design questions for how flexible to be, like accepting *i*/*j* on the imaginary part, or letting the real and/or imaginary parts be optional, etc. But I can't think of anything actually blocking it.

clarcharr:
I would personally prefer parsing from the formats `a+bi` or `a+bj`, allowing for spaces between the plus.

c410-f3r:
[Matrix Market format](http://math.nist.gov/MatrixMarket/formats.html#MMformat) and [Harwell Boeing format](http://people.sc.fsu.edu/~jburkardt/data/hb/hb.html) are ASCII files that use the simplest complex number representation that I just mentioned above.
We could use this representation to implement `Num` and `FromStr` would delegate to `from_str_radix` the same way BigUint does.

aliddell:
I'd be happy to help here. My own preference would be `a+/- bi` or `a +/- bj` but not both; `a`; or `b{i,j}`. It looks like when printed they end in `i`, so for consistency I would vote for that.

clarcharr:
Seems good for me!

cuviper:
Completed in #317.

PR #317, opened 2017-07-10
Feature/complex from str

aliddell:
This commit adds a basic parser for Complex types in Cartesian form, per https://github.com/rust-num/num/issues/289. It will take numbers of the form `a + bi`, `ai + b`, `a - bi`, `ai - b`, `a`, or `ai`. At least one space between the real/imaginary parts and the operator is mandatory; without bringing in a dependency on some regex crate, it's nontrivial to handle cases like, e.g., 0.217828e+1+31.4159E-1, or a similar case with polar coordinates. I could work on these issues later if you like.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -743,12 +743,12 @@ impl<T> fmt::Binary for Complex<T> where
 }
 
 impl<T> FromStr for Complex<T> where
-    T: FromStr + Num + PartialOrd + Clone + std::fmt::Debug, T::Err: Error + PartialEq
+    T: FromStr + Num + PartialOrd + Clone, T::Err: Error
 {
-    type Err = ParseComplexError<T>;
+    type Err = ParseComplexError<T::Err>;
 
     /// Parses `a +/- bi`; `ai +/- b`; `a`; or `bi` where `a` and `b` are of type `T`
-    fn from_str(s: &str) -> Result<Complex<T>, ParseComplexError<T>>
+    fn from_str(s: &str) -> Result<Complex<T>, ParseComplexError<T::Err>>
     {
         let imag = match s.rfind('j') {
             None => 'i',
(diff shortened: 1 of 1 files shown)


clarcharr:
Could you modify this to accept `j` as well?

Also, it looks like the operators require at least one space between them, and it'd be best to accept strings like `"1+2.3i"`.

aliddell:
@clarcharr yeah probably over the next couple days

complex/src/lib.rs
```diff
@@ -740,6 +742,141 @@ impl<T> fmt::Binary for Complex<T> where
...
+        let re = match a.rfind(imag) {
+            None => {
+                try!(T::from_str(&a)
+                    .map_err(|_| ParseComplexError { kind: ComplexErrorKind::ParseError }))
+            },
+            _ => {
+                try!(T::from_str(&b)
+                    .map_err(|_| ParseComplexError { kind: ComplexErrorKind::ParseError }))
+            }
+        };
```
cuviper: I think rather than duplicating these parsing blocks, it would be better to just normalize `a` as real component and `b` as the imaginary beforehand.  Just swap them if needed, then make sure they're not both real or both imaginary, and then parse each piece.

complex/src/lib.rs
```diff
@@ -740,6 +742,141 @@ impl<T> fmt::Binary for Complex<T> where
...
+
+        let im = match a.rfind(imag) {
+            None => {
+                // a is real
+                match b.pop() {
```
cuviper: All the `rfind` should instead be `ends_with`, since you're just popping it off here.
(Although if you normalize the values above, I'd just slice it off then.)

complex/src/lib.rs
```diff
@@ -763,6 +900,36 @@ impl<T> serde::Deserialize for Complex<T> where
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct ParseComplexError {
+    kind: ComplexErrorKind,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum ComplexErrorKind {
```
cuviper: I think this should be parameterized on `T: FromStr`, and have a variant containing `T::Err`.  If you can also have more specific variants for complex's own errors, that would be better.

complex/src/lib.rs
```diff
@@ -178,7 +180,7 @@ impl<T: Clone + Float> Complex<T> {
         let (r, theta) = self.to_polar();
         Complex::from_polar(&(r.sqrt()), &(theta/two))
     }
-    
+
```
cuviper: Please don't include unrelated changes like this.  I won't ask you to revert these this time, because they're few enough that I can look past them, but generally it makes things harder to review.

complex/src/lib.rs
```diff
@@ -1399,4 +1567,78 @@ mod test {
...
+        test(_neg1_1i, "-1+1i".to_string());
+        test(_neg1_1i, "1i-1".to_string());
+        test(_neg1_1i, "j + -1".to_string());
+
+        test(_05_05i, "0.5 + 0.5i".to_string());
+        test(_05_05i, "0.5+0.5j".to_string());
+        test(_05_05i, "5e-1+0.5j".to_string());
+        test(_05_05i, "5E-1 + 0.5j".to_string());
+        test(_05_05i, "5E-1i + 0.5".to_string());
+        test(_05_05i, "0.05e+1j + 50E-2".to_string());
```
cuviper: Lots of test variations, looks nice!

complex/src/lib.rs
```diff
@@ -1399,4 +1567,78 @@ mod test {
         assert!(!b.is_normal());
         assert!(_1_1i.is_normal());
     }
+
+    #[test]
+    fn test_from_str() {
+        fn test(z: Complex64, s: String) {
```
cuviper: If you make this take `&str`, you can avoid all the `to_string()` calls below.

complex/src/lib.rs
aliddell: Sorry about this, I think Atom did these for me without fanfare.

complex/src/lib.rs
```diff
@@ -740,6 +742,141 @@ impl<T> fmt::Binary for Complex<T> where
...
+    /// Parses `a +/- bi`; `ai +/- b`; `a`; or `bi` where `a` and `b` are of type `T`
+    fn from_str(s: &str) -> Result<Complex<T>, ParseComplexError>
+    {
+        let imag = match s.rfind('j') {
+            None => 'i',
+            _ => 'j'
+        };
+
+        let mut a = String::with_capacity(s.len());
+        let mut b = String::with_capacity(s.len());
```
aliddell: IIRC this is an artifact from my first attempt at a parser. We could probably end up reusing `s` for the real part and just allocate for the imaginary part. Is that what you're getting at?

aliddell:
@cuviper Thanks for all the feedback! Where stuff's not clear to me I've asked questions, but otherwise I can get on these commits later today.

complex/src/lib.rs
aliddell: This is not as straightforward as I thought it would be. I tried to parameterize on `T: FromStr` and ended up having to add a few traits to ComplexErrorKind. I'm still pretty new to Rust, so let me know if these are reasonable or if I'm missing anything.

complex/src/lib.rs
```diff
@@ -740,6 +742,107 @@ impl<T> fmt::Binary for Complex<T> where
...
+        let mut split_index = s.len();
+
+        loop {
+            match char_indices.next() {
+                Some(t) => {
+                    let i = t.0;
+                    let cc = t.1;
+
+                    if cc == '+' && pc != 'e' && pc != 'E' {
+                        // ignore '+' if part of an exponent
```
cuviper: Rust parses `+` at the beginning of numbers too, so we should allow and include that.  But I'd say still skip the middle `+`, so we don't have to *force* the explicit sign.

cuviper:
Thanks!  Hope you don't mind I made a few final tweaks myself.  Let's see what the bot thinks...

bors r+

PR #325, opened 2017-08-02
Implement assign_from_slice

kompass:
This commit implements `assign_from_slice(..)` for `BigUint` and `BigInt`. `assig_from_slice` is for performance : it allows to reassign a value to a `BigInt` or a `BigUint` without useless reallocations. It would be useful for loops for example. I need it to implement the Toom-3 algorithm.

I also added a missing test for `BigInt::from_slice(..)`.

### DIFF
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -1395,6 +1395,21 @@ impl BigInt {
         BigInt::from_biguint(sign, BigUint::from_slice(slice))
     }
 
+    /// Reinitializes a `BigInt`.
+    #[inline]
+    pub fn assign_from_slice(&mut self, sign: Sign, slice: &[BigDigit]) {
+        // Normalize:
+        let slice = &slice[..slice.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1)];
+        
+        if sign == NoSign || slice.len() == 0 {
+            self.sign = NoSign;
+            self.data = Zero::zero();
+        } else {
+            self.sign = sign;
+            self.data.assign_from_slice(slice);
+        }
+    }
+
     /// Creates and initializes a `BigInt`.
     ///
     /// The bytes are in big-endian byte order.
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -1376,13 +1376,13 @@ impl BigInt {
     ///
     /// The digits are in little-endian base 2^32.
     #[inline]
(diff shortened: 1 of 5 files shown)


bigint/src/bigint.rs
```diff
@@ -1395,6 +1395,21 @@ impl BigInt {
...
 
+    /// Reinitializes a `BigInt`.
+    #[inline]
+    pub fn assign_from_slice(&mut self, sign: Sign, slice: &[BigDigit]) {
+        // Normalize:
+        let slice = &slice[..slice.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1)];
+        
+        if sign == NoSign || slice.len() == 0 {
+            self.sign = NoSign;
+            self.data = Zero::zero();
```
cuviper: If `assign_from_slice` aims to preserve the allocation, this line is throwing that away.  You could use `self.data.assign_from_slice(&[])` instead.

(Hmm, I wonder if `from_biguint` should do that too...)

bigint/src/bigint.rs
```diff
@@ -1395,6 +1395,21 @@ impl BigInt {
         BigInt::from_biguint(sign, BigUint::from_slice(slice))
     }
 
+    /// Reinitializes a `BigInt`.
+    #[inline]
+    pub fn assign_from_slice(&mut self, sign: Sign, slice: &[BigDigit]) {
+        // Normalize:
+        let slice = &slice[..slice.iter().rposition(|&x| x != 0).map_or(0, |i| i + 1)];
```
kompass: Yes you're totally right, no need to normalize twice, I will fix it.

bigint/src/bigint.rs
kompass: Yes I will fix it, and optimize `from_biguint` too.

kompass:
About the name, since copying and cloning are the same things for scalars, we can rename `assign_from_slice` to `copy_from_slice` even though we use `Vec::clone_from_slice`. The name of the method doesn't matter to me as long as it is consistent, so as you want.

kompass:
The current implementation of Karatsuba allocates one Vec `p` to each function call. Since the calls are recursive, we could consider reusing this Vec, but `p` is already passed as a parameter so currently, I don't think this is possible. I will look for a way after implementing Toom-3.

For Toom-3, since the algorithm use more intermediate values (13), I use assign_from_slice to reduce the number of Vecs allocated to each call.

cuviper:
Alright, let's just keep the name.  Thanks!

bors r+

Issue #321, opened 2017-07-18
Need common trait for floating and complex numbers


carrutstick:
This issue in in the vein of issue #209, but a bit more ambitious; hopefully I am not too far out of scope for this project.

Any mathematical function that accepts a real number can technically also accept a complex number, but currently there is no trait we can use to define a function that is generic over real and complex numbers (unless I'm missing something). 

I see two possible solutions:
1. Shoehorn complex numbers into the existing `Float` trait (minimal work)
2. Refactor the float trait so that we separate the parts that refer to mathematical objects (the trig functions, for instance) from the parts that assume specific types of implementation (`min_value()`, `max()`, etc.).

Option 1
-----------

Certain functions in the `Float` trait may be ambiguous for complex numbers, such as the `min_value()` function. Broadly, there are three classes of functions in the trait:

1. `|Self| -> Self`
2. `|| -> Self`
3. `|Self| -> bool`

Obviously the trigonometric functions of class 1 are not ambiguous (once we pick a branch cut), but functions like `floor()` and `frac()` are less obvious. I think a good rule of thumb for such functions is to have them apply independently to each part of the complex number, so that
```
Complex::new(x, y).floor() == Complex::new(x.floor(), y.floor())
```

For class 2, I'm much less certain, but I think that these should generally produce values that are mathematically equivalent to their real counterparts, so that
```
Complex<F>::min_value() == Complex<F>::new(F::MIN, F::zero())
Complex<F>::infinity() == Complex<F>::new(F::infinity(), F::zero())
```

For class 3 I don't think there's a good rule of thumb, and we would probably need to invent rules on a case-by-case basis. For instance, we would probably want a complex number to have `is_nan() == true` if either part was `nan`, but what if the other part is infinite? Do we also have `is_infinite() == true`? Or do we decide that something that is `nan` cannot be infinite?

There is also the special case of the `min` and `max` functions. I think the only consistent way to handle these is to just have them act on the real line, like class 2.

Option 2
-----------

[...]

Issue #209, opened 2016-07-21
Compound types, e.g. Complex<f64> do not implement the Num trait


beneills:
I have been trying to create matricies with elements of type `Complex<f64>`.

I'm using the `la::Matrix` struct, but this library requires the element type to implement the `Num` trait.  This is not implemented for `Complex<f64>`.  Adding this impl to my code fails, and trying to incorporate it into a forked num crate seems non-trivial, since putting it in either _traits/_ or _complex/_ results in dependency issues.
# Reproductiion Steps

```
let m: Matrix<Complex<f64>> = Matrix::id(width, width);
error: the trait bound `num::Complex<f64>: num::Num` is not satisfied [E0277]
```

cuviper:
You'll need to implement `Rem` and `Num`'s `fn from_str_radix`.

For the dependencies, it can only go in the `num-complex` crate.  What problem did you have with that?

termoshtt:
I have encountered similar problem.
I am trying to implement `from_str_radix` for Complex64 using implementation for `f64`.

But I found a comment in impl macro of `Num`:
https://github.com/rust-num/num/blob/master/traits/src/lib.rs#L89

> FIXME: The standard library from_str_radix on floats was deprecated, so we're stuck
with this implementation ourselves until we want to make a breaking change.
(would have to drop it from `Num` though)

I hope that `from_str_radix` is splited from `Num` into another trait.

cuviper:
`Complex` does implement `Num` since #327.

PR #327, opened 2017-08-09
Add general Rem and Num implementations for Complex<T>

carrutstick:
This should address #209 with eyes towards addressing #321.

It was a little tricky to get `Rem` working for a general `Num`, and I had to add a `PartialOrd` constraint to get it working, but I think it should be fairly robust.

I could probably use extra eyes on the `from_str_radix` function, as I mostly lifted the code from the `from_str` function and I may be missing some subtleties in how that works.

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -531,30 +531,7 @@ impl<T: Clone + Num + PartialOrd> Rem<Complex<T>> for Complex<T> {
         // This is the gaussian integer corresponding to the true ratio
         // rounded towards zero.
         let (re0, im0) = (re.clone() - re % T::one(), im.clone() - im % T::one());
-
-        let zero = T::zero();
-        let one  = T::one();
-        let neg  = zero.clone() - one.clone();
-        // Traverse the 3x3 square of gaussian integers surrounding our
-        // current approximation, and select the one whose product with
-        // `modulus` is closest to `self`.
-        let mut bestrem = self.clone() - modulus.clone() *
-                          Complex::new(re0.clone(), im0.clone());
-        let mut bestnorm = bestrem.norm_sqr();
-        for &(dr, di) in
-            vec![(&one, &zero), (&one, &one), (&zero, &one), (&neg, &one),
-                 (&neg, &zero), (&neg, &neg), (&zero, &neg), (&one, &neg)]
-                            .iter() {
-            let newrem = self.clone() - modulus.clone() *
-                         Complex::new(re0.clone() + dr.clone(),
-                                      im0.clone() + di.clone());
-            let newnorm = newrem.norm_sqr();
-            if newnorm < bestnorm {
-                bestrem = newrem;
-                bestnorm = newnorm;
-            }
-        }
-        bestrem
+        self - modulus * Complex::new(re0, im0)
     }
 }
 
@@ -1702,7 +1679,7 @@ mod test {
 
     mod real_arithmetic {
         use super::super::Complex;
-        use super::_4_2i;
+        use super::{_4_2i, _neg1_1i};
 
         #[test]
         fn test_add() {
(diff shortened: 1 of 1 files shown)


cuviper:
Just FYI it's going to be a week or so before I can review this in depth.

complex/src/lib.rs
```diff
@@ -970,6 +1054,95 @@ impl<T> FromStr for Complex<T> where
     }
 }
 
+impl<T: Num + Clone + PartialOrd> Num for Complex<T> {
+    type FromStrRadixErr = ParseComplexError<T::FromStrRadixErr>;
+
+    /// Parses `a +/- bi`; `ai +/- b`; `a`; or `bi` where `a` and `b` are of type `T`
+    fn from_str_radix(s: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr>
```
cuviper: This is identical to the `from_str` implementation, apart from nested calls to `from-str_radix` instead, right?  Let's see if we can use a common helper function instead, something like:

```rust
fn from_str_any<T, E, F: Fn(&str) -> Result<T, E>>(s: &str, from: F) -> Result<Complex<T>, E> { ... }
```

- `from_str` calls `from_str_any(s, T::from_str)`
- `from_str_radix` calls `from_str_any(s, |s| T::from_str_radix(s, radix))`

complex/src/lib.rs
```diff
@@ -518,10 +518,50 @@ impl<T: Clone + Num> Div<Complex<T>> for Complex<T> {
...
+        // This is the gaussian integer corresponding to the true ratio
+        // rounded towards zero.
+        let (re0, im0) = (re.clone() - re % T::one(), im.clone() - im % T::one());
+
+        let zero = T::zero();
+        let one  = T::one();
+        let neg  = zero.clone() - one.clone();
+        // Traverse the 3x3 square of gaussian integers surrounding our
+        // current approximation, and select the one whose product with
+        // `modulus` is closest to `self`.
```
cuviper: So you're basically rounding the division to the nearest gaussian?  I think this is makes it overly complicated, whereas simply rounding towards zero will at least agree with pure-real `Rem`.

e.g. `2.0 % 3.0 == 2.0`, but in `Complex` your implementation gives `-1.0`.

carrutstick: The funny thing about rounding towards zero is that you can end up with remainders that have larger magnitude than your modulus. For instance you'd have (3 + 3i) % 4 == 3 + 3i, which has greater magnitude than 4.

Of course if we go with lexicographical ordering of complex numbers, similar issues might be unavoidable. I'm not sure there is any sensible scheme where the remainder is never lexicographically greater than the modulus.

cuviper: Yes, that is a funny example.  Do you have one that would fail lexicographical ordering?

I guess we see why ordering and remainders aren't usually implemented for complex numbers.  But if we do implement them, I think it's a fundamental property that operations on `Complex` values with zero imaginary should match the equivalent operation on the real part alone.

carrutstick: So one example is (1+5i) % 2i: your quotient is 2.5-.5i, which rounds to 2, and then your remainder is 1+1i, which is lexicographically larger than your modulus.

Are you aware of any compiler warnings along the line of `#[deprecated]` that would be appropriate for functions that may have unexpected behavior? And what would you think of applying them to `rem` and `partial_cmp`?

carrutstick: P.S. I thing having complex numbers with zero imaginary component try to match behavior with real numbers is a fine principle; just pointing out places where it can get weird.

complex/src/lib.rs
```diff
@@ -758,6 +810,15 @@ impl<T: Clone + Num> Div<T> for Complex<T> {
     }
 }
 
+impl<T: Clone + Num + PartialOrd> Rem<T> for Complex<T> {
```
cuviper: I think now this doesn't need `PartialOrd`, right?  And we can revert the related macro changes.

carrutstick: Yes, good point.

PR #329, opened 2017-08-25
Implement Display for num_bigint::Sign

FliegendeWurst:


bigint/src/bigint.rs
```diff
@@ -101,6 +101,17 @@ impl serde::Deserialize for Sign {
...
 }
 
+impl fmt::Display for Sign {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		use Sign::*;
+		match *self {
+			Plus => write!(f, "+"),
+			NoSign => write!(f, ""),
+			Minus => write!(f, "-")
+		}
```
cuviper: I think using [`Formatter::pad(&mut self, s: &str)`](https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.pad) will be more flexible than `write!`.

FliegendeWurst: OK :)

FliegendeWurst:
Here's an [example](https://play.rust-lang.org/?gist=a8b80c748f86400f25ffde35dfba3ba7).
Tests have been added

cuviper:
Did you know you can get almost the same effect with "{:+}" format?  Except that a zero will be printed as "+0", just like primitive integers do, unlike what you've done for `NoSign`.

So your example shows *how* to use it, but I'm looking for *why* you need it.

FliegendeWurst:
Oh I don't need it in that case, thanks (you may close this PR if you want)

cuviper:
OK, thanks anyway!

PR #334, opened 2017-09-21
Use `Mul` by reference in `pow`

cuviper:
Resubmission of #153, since it wasn't open to maintainer edits.  I added a distinct `Float::powi` fallback since that doesn't have `Mul` by reference.

cc @dten

### DIFF
--- a/traits/src/float.rs
+++ b/traits/src/float.rs
@@ -459,13 +459,33 @@ pub trait Float
     /// assert!(abs_difference < 1e-10);
     /// ```
     #[inline]
-    fn powi(mut self, mut exp: i32) -> Self {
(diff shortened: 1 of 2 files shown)


dten:
Shame i lost my commit contribute but oh well. Thanks

cuviper:
Your commit still has you as the author, just me as the committer since I rebased it.  Is that what you mean?

dten:
Doesn't show in contributors for some reason. No idea how github makes those stats. Doesn't matter i just like numbers against my name haha

cuviper:
Oh, I think that's just because it's not on master. :)

dten:
Ahh. The wait begins

PR #338, opened 2017-09-29
Scalar trait: common trait for floating and complex numbers

termoshtt:
Related to #321 (or resolve it?)

I introduce a trait `Scalar` to use `Complex` and `Float` transparently.
Most of numerical functions (e.g. `sin`) are defined both in the `Float` trait and `Complex`. It is not useful to implement an algorithm which can accept both floating and complex numbers (e.g. most of matrix manipulation in [ndarray-linalg](https://github.com/termoshtt/ndarray-linalg))

### DIFF
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -22,7 +22,8 @@ extern crate rustc_serialize;
 #[cfg(feature = "serde")]
 extern crate serde;
 
-pub mod scalar;
+mod scalar;
+pub use scalar::Scalar;
 
 use std::error::Error;
 use std::fmt;
--- a/complex/src/lib.rs
+++ b/complex/src/lib.rs
@@ -22,6 +22,8 @@ extern crate rustc_serialize;
 #[cfg(feature = "serde")]
 extern crate serde;
 
+pub mod scalar;
+
 use std::error::Error;
 use std::fmt;
 #[cfg(test)]
--- a/complex/src/scalar.rs
+++ b/complex/src/scalar.rs
@@ -4,9 +4,22 @@ use std::ops::Neg;
 
 use Complex;
 
-pub trait Scalar: Num + Copy + Neg<Output = Self> {
-    /// Associated Real type
-    type Real: Scalar<Real = Self::Real>;
+pub trait Associated {
+    type Real: Scalar;
+    type Complex: Scalar;
+}
+
+impl<T: Scalar> Associated for T
+where
+    Complex<T::Repr>: Scalar,
+{
+    type Real = T::Repr;
+    type Complex = Complex<T::Repr>;
+}
+
+pub trait Scalar: Num + Copy + Neg<Output = Self> + Associated {
+    /// Associated Repr type
+    type Repr: Scalar<Repr = Self::Repr>;
 
     /// Take the square root of a number.
     fn sqrt(&self) -> Self;
@@ -18,7 +31,7 @@ pub trait Scalar: Num + Copy + Neg<Output = Self> {
     fn ln(&self) -> Self;
 
     /// Returns the square of the absolute value of the number
-    fn abs_sqr(&self) -> Self::Real;
+    fn abs_sqr(&self) -> Self::Repr;
 
     /// Returns the absolute value of the number
     fn abs(&self) -> Self::Real;
@@ -27,10 +40,10 @@ pub trait Scalar: Num + Copy + Neg<Output = Self> {
     fn powi(&self, exp: i32) -> Self;
 
     /// Raise a number to a floating point power.
-    fn powf(&self, exp: Self::Real) -> Self;
+    fn powf(&self, exp: Self::Repr) -> Self;
 
     /// Raise a number to a complex power.
-    fn powc(&self, exp: Complex<Self::Real>) -> Complex<Self::Real>;
+    fn powc(&self, exp: Self::Complex) -> Complex<Self::Repr>;
 
     /// Returns complex-conjugate number
     fn conj(&self) -> Self;
@@ -85,7 +98,7 @@ pub trait Scalar: Num + Copy + Neg<Output = Self> {
 }
 
 impl<T: Clone + Float + FromPrimitive> Scalar for Complex<T> {
-    type Real = T;
+    type Repr = T;
 
     fn sqrt(&self) -> Self {
         Complex::sqrt(self)
@@ -116,7 +129,7 @@ impl<T: Clone + Float + FromPrimitive> Scalar for Complex<T> {
         Complex::powf(self, exp)
     }
 
-    fn powc(&self, exp: Complex<Self::Real>) -> Complex<Self::Real> {
+    fn powc(&self, exp: Self::Complex) -> Self::Complex {
         Complex::powc(self, exp)
     }
 
@@ -189,8 +202,11 @@ impl<T: Clone + Float + FromPrimitive> Scalar for Complex<T> {
     }
 }
 
-impl<T: Float> Scalar for T {
-    type Real = T;
+impl<T: Float> Scalar for T
+where
+    Complex<T>: Scalar,
+{
+    type Repr = T;
 
     fn sqrt(&self) -> Self {
         Float::sqrt(*self)
@@ -220,7 +236,7 @@ impl<T: Float> Scalar for T {
         Float::powf(*self, exp)
     }
 
-    fn powc(&self, exp: Complex<Self::Real>) -> Complex<Self::Real> {
+    fn powc(&self, exp: Self::Complex) -> Self::Complex {
         exp.expf(*self)
     }
 
--- a/complex/src/scalar.rs
+++ b/complex/src/scalar.rs
@@ -1,0 +1,121 @@
+
+use traits::{Num, Float};
+use std::ops::Neg;
+
+use Complex;
+
+pub trait Scalar: Num + Copy + Neg<Output = Self> {
+    /// Associated Real type
+    type Real;
+    /// Associated Complex type
+    type Complex;
+
+    fn sqrt(&self) -> Self;
+    fn exp(&self) -> Self;
+    fn ln(&self) -> Self;
+    fn abs_sqr(&self) -> Self::Real;
+    fn abs(&self) -> Self::Real;
+
+    fn conj(&self) -> Self;
+
+    // trigonometric functions
+    fn sin(&self) -> Self;
+    fn cos(&self) -> Self;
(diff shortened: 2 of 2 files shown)


complex/src/lib.rs
```diff
@@ -22,6 +22,8 @@ extern crate rustc_serialize;
 #[cfg(feature = "serde")]
 extern crate serde;
 
+pub mod scalar;
```
cuviper: It's good to implement this in a module, but that shouldn't be public.  Just `pub use scalar::Scalar` to expose it.

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
...
+pub trait Scalar: Num + Copy + Neg<Output = Self> {
+    /// Associated Real type
+    type Real;
+    /// Associated Complex type
+    type Complex;
+
+    /// Take the square root of a number.
+    fn sqrt(&self) -> Self;
+    /// Returns `e^(self)`, (the exponential function).
+    fn exp(&self) -> Self;
```
cuviper: nit: please put a blank line between all `type`/`fn` items.  If you want visual clustering of related items, you can use two blank lines around them.

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
+
+use traits::{Num, FromPrimitive, Float};
+use std::ops::Neg;
+
+use Complex;
+
+pub trait Scalar: Num + Copy + Neg<Output = Self> {
+    /// Associated Real type
+    type Real;
```
cuviper: Should we have a constraint like `Real: Scalar`?

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Associated Complex type
+    type Complex;
```
cuviper: Do we really need this flexibility?  Perhaps we should just use `Complex<Self::Real>` where needed?  That's only for `powc`, so I wonder if that's worth its weight here at all.

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
...
+        Float::sqrt(*self)
+    }
+    fn exp(&self) -> Self {
+        Float::exp(*self)
+    }
+    fn ln(&self) -> Self {
+        Float::ln(*self)
+    }
+    fn abs_sqr(&self) -> Self::Real {
+        Float::abs(*self).powi(2)
```
cuviper: `self * self` is probably more efficient.

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
...
+    // powers
+    fn powi(&self, exp: i32) -> Self {
+        Float::powi(*self, exp)
+    }
+    fn powf(&self, exp: Self::Real) -> Self {
+        Float::powf(*self, exp)
+    }
+    fn powc(&self, exp: Self::Complex) -> Self::Complex {
+        let c = Complex::new(*self, T::zero());
+        c.powc(exp)
```
cuviper: `Complex::expf` should help here.

complex/src/scalar.rs
```diff
@@ -0,0 +1,252 @@
...
+    fn is_nan(self) -> bool;
+    /// Checks if the given (real or imaginary part of complex) number is infinite
+    fn is_infinite(self) -> bool;
+    /// Checks if the given number is finite
+    fn is_finite(self) -> bool;
+    /// Checks if the given number is normal
+    fn is_normal(self) -> bool;
+}
+
+impl<T: Clone + Float + FromPrimitive> Scalar for Complex<T> {
```
cuviper: Hmm, it's a little awkward that `FromPrimitive` is needed just for `powi`.

termoshtt:
Thanks for your comment and review!

> In particular, for the methods using Self::Real and Self::Complex, how useful is this actually in code that would be generic over any Scalar? I feel like it will be hard to actually call those methods without locking them into particular types. Can you offer any non-trivial examples where this is helpful?

`Self::Real` and `Self::Complex` is most important in this PR. I introduced it in order to implement an eigenvalue decomposition function (not implemented yet :<). The eigenvalue of both real and complex matrices are complex value:

```rust
fn eigen_values<T: Scalar>(a: Matrix<T>) -> Vec<T::Complex> { ... }
```

Similar thing happens for Hermite matrices which has real eigenvalues even if the components of the matrix is complex:

```rust
fn eigen_values_hermite<T: Scalar>(a: Matrix<T> -> Vec<T::Real> { ... }
```

> Do we really need this flexibility? Perhaps we should just use Complex<Self::Real> where needed?

As you say in the comment, `Scalar::Complex` should be always `Complex<Scalar::Real>`, and I have removed it. If allowed, I would like to keep it for easy typing.

complex/src/scalar.rs
```diff
@@ -0,0 +1,294 @@
+
+use traits::{Num, FromPrimitive, Float};
+use std::ops::Neg;
+
+use Complex;
+
+pub trait Scalar: Num + Copy + Neg<Output = Self> {
+    /// Associated Real type
+    type Real: Scalar<Real = Self::Real>;
```
cuviper: I like the self-referential constraint!  It seems like the kind of thing the compiler might not be able to untangle, but it looks like that worked out.

If we add the associated `type Complex` back, can it be nailed down to the concrete `Complex` type?  Something like:

```rust
type Real: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>;
type Complex: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>;
```

cuviper: When I think more about it, I'm torn whether we should allow other `Complex` types or not.  In theory, someone could want to implement their own type, maybe even just a newtype wrapper.  Then these self-referential constraints would be a burden, e.g. `MyComplex<f64>` would also have to wrap the `Real` type to meet the constraits, since with `Real = f64`, `f64::Complex` would not be `MyComplex`.

What do you think?

termoshtt:
> If we add the associated type Complex back, can it be nailed down to the concrete Complex type? 

Your code does not works:

```
the trait `scalar::Scalar` is not implemented for `Complex<<Self as scalar::Scalar>::Real>`
```

I cannot resolve this error blocked by a recurrent trait dependency, but I found another way by splitting the `Real` and `Complex` to another trait `Associated` https://github.com/rust-num/num/pull/338/commits/fa50a65ea97fcffd323faa684a1f64ef7405d479#diff-2d36ec43057b3c9a6b1d35a5852450d9R7
Corresponding this split, I rename `Scalar::Real` to `Repr` (representation).
This reminds me the CRTP in C++ :<, but it works well :)

> I'm torn whether we should allow other Complex types or not. 

IMO, we should not allow other `Complex` since it is implemented on `num-complex` instead of `num-traits`. If one introduces another `Complex` type, he should define another `Scalar` like this.

complex/src/scalar.rs
termoshtt: In my experience using `ndarray_linalg::Scalar`, the original of this PR, `Scalar::from_f64` is necessary.
I would like to add `trait Scalar: FromPrimitive + ...` restriction, but `Complex` does not implement it.

cuviper:
I'd rather not introduce a side trait -- that becomes yet another thing that users must import if they want to reach `T::Real` and `T::Complex`.  And you didn't re-export it to the root module anyway, so I doubt - can it be seen at all?

The suggestion I made does appear to work if you add all of the `T` constraints to `Real` too, e.g.

```rust
pub trait Scalar: [...] {
    type Real: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>
        + Clone + Float + FromPrimitive;
    type Complex: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>;

    [...]
}
```

It's a bit ugly, but I think it's OK since users don't have to worry about it.

termoshtt:
> The suggestion I made does appear to work if you add all of the T constraints to Real too, e.g.

I succeeds to compile it!

cuviper:
Unfortunately, using a `where` clause falls afoul of rust-lang/rust#20671.  I tried adding this test:

```rust
#[test]
fn scalar_sqrt() {
    fn call_sqrt<T: Scalar>(x: T) -> T {
        x.sqrt()
    }

    call_sqrt(2.0_f64);
    call_sqrt(Complex::from(2.0_f64));
}
```

```text
error[E0277]: the trait bound `Complex<<T as scalar::Scalar>::Real>: scalar::Scalar` is not satisfied
   --> src/scalar.rs:309:5
    |
309 | /     fn call_sqrt<T: Scalar>(x: T) -> T {
310 | |         x.sqrt()
311 | |     }
    | |_____^ the trait `scalar::Scalar` is not implemented for `Complex<<T as scalar::Scalar>::Real>`
    |
    = help: consider adding a `where Complex<<T as scalar::Scalar>::Real>: scalar::Scalar` bound
    = note: required by `scalar::Scalar`
```

But it does work if I drop that `where` clause and add the constraint to `Real` as I mentioned before:

```rust
    /// Associated Real type
    type Real: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>
        + Float + FromPrimitive;
```

Please do add some `Scalar` tests so we can make sure expected use-cases work!

cuviper:
Unfortunately, using a `where` clause falls afoul of rust-lang/rust#20671.  I tried adding this test:

```rust
#[test]
fn scalar_sqrt() {
    fn call_sqrt<T: Scalar>(x: T) -> T {
        x.sqrt()
    }

    call_sqrt(2.0_f64);
    call_sqrt(Complex::from(2.0_f64));
}
```

```text
error[E0277]: the trait bound `Complex<<T as scalar::Scalar>::Real>: scalar::Scalar` is not satisfied
   --> src/scalar.rs:309:5
    |
309 | /     fn call_sqrt<T: Scalar>(x: T) -> T {
310 | |         x.sqrt()
311 | |     }
    | |_____^ the trait `scalar::Scalar` is not implemented for `Complex<<T as scalar::Scalar>::Real>`
    |
    = help: consider adding a `where Complex<<T as scalar::Scalar>::Real>: scalar::Scalar` bound
    = note: required by `scalar::Scalar`
```

But it does work if I drop that `where` clause and add the constraint to `Real` as I mentioned before:

```rust
    /// Associated Real type
    type Real: Scalar<Real = Self::Real, Complex = Complex<Self::Real>>
        + Float + FromPrimitive;
```

Please do add some `Scalar` tests so we can make sure expected use-cases work!

cuviper:
Please see the new repo and issue: rust-num/num-complex#2

I'm still interested in this PR, if you'd like to move it over.

PR #339, opened 2017-10-07
Implement modpow() for BigUint backed by Montgomery Multiplication

### LINKED ISSUE #136: Montgomery modular multiplication
I implemented the montgomery modular multiplication for BigUint, and it produces a pretty nice speedup (6x) for my modular exponentiation implementation.

You can find the code snippet [here](https://gist.github.com/yshui/027eecdf95248ea69606)

I think this would be a nice addition to rust-num

str4d:
Based on this Gist: https://gist.github.com/yshui/027eecdf95248ea69606

Also adds support to `BigUint.from_str_radix()` for using `_` as a visual separator.

Closes #136

### DIFF
--- a/benches/bigint.rs
+++ b/benches/bigint.rs
@@ -256,25 +256,39 @@ fn pow_bench(b: &mut Bencher) {
     });
 }
 
+
+/// This modulus is the prime from the 2048-bit MODP DH group:
+/// https://tools.ietf.org/html/rfc3526#section-3
+const RFC3526_2048BIT_MODP_GROUP: &'static str = "\
+    FFFFFFFF_FFFFFFFF_C90FDAA2_2168C234_C4C6628B_80DC1CD1\
+    29024E08_8A67CC74_020BBEA6_3B139B22_514A0879_8E3404DD\
+    EF9519B3_CD3A431B_302B0A6D_F25F1437_4FE1356D_6D51C245\
+    E485B576_625E7EC6_F44C42E9_A637ED6B_0BFF5CB6_F406B7ED\
+    EE386BFB_5A899FA5_AE9F2411_7C4B1FE6_49286651_ECE45B3D\
+    C2007CB8_A163BF05_98DA4836_1C55D39A_69163FA8_FD24CF5F\
+    83655D23_DCA3AD96_1C62F356_208552BB_9ED52907_7096966D\
+    670C354E_4ABC9804_F1746C08_CA18217C_32905E46_2E36CE3B\
+    E39E772C_180E8603_9B2783A2_EC07A28F_B5C55DF0_6F4C52C9\
+    DE2BCBF6_95581718_3995497C_EA956AE5_15D22618_98FA0510\
+    15728E5A_8AACAA68_FFFFFFFF_FFFFFFFF";
+
 #[bench]
 fn modpow(b: &mut Bencher) {
     let mut rng = get_rng();
     let base = rng.gen_biguint(2048);
     let e = rng.gen_biguint(2048);
-    // This modulus is the prime from the 2048-bit MODP DH group:
-    // https://tools.ietf.org/html/rfc3526#section-3
-    let m = BigUint::from_str_radix("\
-        FFFFFFFF_FFFFFFFF_C90FDAA2_2168C234_C4C6628B_80DC1CD1\
-        29024E08_8A67CC74_020BBEA6_3B139B22_514A0879_8E3404DD\
-        EF9519B3_CD3A431B_302B0A6D_F25F1437_4FE1356D_6D51C245\
-        E485B576_625E7EC6_F44C42E9_A637ED6B_0BFF5CB6_F406B7ED\
-        EE386BFB_5A899FA5_AE9F2411_7C4B1FE6_49286651_ECE45B3D\
-        C2007CB8_A163BF05_98DA4836_1C55D39A_69163FA8_FD24CF5F\
-        83655D23_DCA3AD96_1C62F356_208552BB_9ED52907_7096966D\
-        670C354E_4ABC9804_F1746C08_CA18217C_32905E46_2E36CE3B\
-        E39E772C_180E8603_9B2783A2_EC07A28F_B5C55DF0_6F4C52C9\
-        DE2BCBF6_95581718_3995497C_EA956AE5_15D22618_98FA0510\
-        15728E5A_8AACAA68_FFFFFFFF_FFFFFFFF", 16).unwrap();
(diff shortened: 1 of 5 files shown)


str4d:
I'm getting strange runtimes with this PR. The benchmarks indicate that for 2048-bit bases and exponents, the runtime is reasonable:

```
$ rustup run nightly cargo bench --bench bigint modpow
    Finished release [optimized] target(s) in 0.0 secs
     Running target/release/deps/bigint-7cd7b2fb23889fda

running 1 test
test modpow            ... bench:  38,196,475 ns/iter (+/- 6,887,198)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured; 30 filtered out
```

But when I use `ModPow` in my own code (or for that matter when I instrument the `modpow_big` test in this PR), I get significantly longer times (with no difference between stable and nightly):

```
SKBuilder modpow: Duration { secs: 3, nanos: 314462408 }
```

benches/bigint.rs
```diff
@@ -255,3 +255,39 @@ fn pow_bench(b: &mut Bencher) {
         }
     });
 }
+
+#[bench]
+fn modpow(b: &mut Bencher) {
+    let mut rng = get_rng();
+    let base = rng.gen_biguint(2048);
+    let e = rng.gen_biguint(2048);
+    let m =
```
str4d: The modulus used here and in the tests is the prime from the [2048-bit MODP DH group](https://tools.ietf.org/html/rfc3526#section-3) (used because I was testing this against I2P's DH agreement).

str4d:
> I'm getting strange runtimes with this PR.

Never mind, it was pointed out to me that `--release` is set by default for benchmarks but not for anything else.

benches/bigint.rs
cuviper: That info belongs in a direct code comment!  And perhaps it will be clearer to parse it from a string in the same grouping as given in that source link, e.g. `"0xFFFFFFFF_FFFFFFFF_C90FDAA2_2168C234_..."`.

(Hmm, do we not parse `_` in bigints?  We should!)

bigint/src/algorithms.rs
```diff
@@ -555,6 +556,124 @@ pub fn ilog2<T: traits::PrimInt>(v: T) -> usize {
     fls(v) - 1
 }
 
+pub struct MontyReducer<'a> {
```
cuviper: Let's put all this in its own `monty` module, instead of adding to `algorithms`.

Also, I realize it wasn't your code to begin with, but if you can figure out some good code comments to add to this implementation, it would help a lot for reviewing and future maintenance.

bigint/src/biguint.rs
```diff
@@ -976,6 +977,17 @@ impl Integer for BigUint {
     }
 }
 
+pub trait ModPow {
+    fn modpow(&self, exponent: &Self, modulus: &Self) -> Self;
```
cuviper: Why is this a trait instead of just adding a method to `BigUint`?

If we really want a trait, it would belong in `num-traits`, hopefully implemented for primitive integers too.  But I think that's unnecessary for the moment.

bigint/src/algorithms.rs
```diff
@@ -555,6 +556,124 @@ pub fn ilog2<T: traits::PrimInt>(v: T) -> usize {
...
+
+// Montgomery Multiplication
+fn monty_mult(a: BigUint, b: &BigUint, mr: &MontyReducer) -> BigUint {
+    monty_redc(a * b, mr)
+}
+
+// Montgomery Squaring
+fn monty_sqr(a: BigUint, mr: &MontyReducer) -> BigUint {
+    // TODO: Replace pow with an optimised squaring function
+    monty_redc(pow(a, 2), mr)
```
cuviper: Curious, even a simple `&a * &a` should be better than `pow` for squaring.

benches/bigint.rs
cuviper: (It seems primitive integers don't parse `_` either.  I'm surprised, since it is allowed in integer literals.)

bigint/src/algorithms.rs
str4d: ACK. And yes, I planned to add comments before making the PR, but the strange (self-inflicted) behaviour motivated me to get early feedback :)

bigint/src/algorithms.rs
str4d: Didn't think to change it from what the original code had, but yes it would be. Not that it will affect performance much, as `pow(a, 2)` amounts to a wrapper around `& * &a`.

bigint/src/biguint.rs
str4d: I made a trait because that was how every other mathematical operator was implemented on `BigUint`, so I was trying to follow precedent. I'll move it into `impl BigUint`.

str4d:
@cuviper addressed your comments.

bigint/src/tests/biguint.rs
```diff
@@ -1089,6 +1089,79 @@ fn test_is_even() {
     assert!(((&one << 64) + one).is_odd());
 }
 
+#[test]
+fn test_modpow() {
+    fn check(b: usize, e: usize, m: usize, r: usize) {
+        let big_b: BigUint = FromPrimitive::from_usize(b).unwrap();
```
cuviper: `BigUint::from` is infallible.

bigint/src/biguint.rs
```diff
@@ -1611,6 +1615,12 @@ impl BigUint {
         self.normalize();
         self
     }
+
+    /// Returns `(self ^ exponent) % modulus`.
```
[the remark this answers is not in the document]
cuviper: Actually, it shouldn't be too bad to just have a squaring loop until the exponent is odd, then use monty from there.

bigint/src/monty.rs
```diff
@@ -0,0 +1,155 @@
...
+    if u < 0 {
+        (u + mu) as u64
+    } else {
+        u as u64
+    }
+}
+
+impl<'a> MontyReducer<'a> {
+    pub fn new(p: &'a BigUint) -> Self {
+        let n : Vec<u32> = p.data.clone();
```
cuviper: I think we don't actually need this cloned `n` -- since we also save `p`, we can just access `mr.p.data`.

cuviper:
I pushed a bunch of changes myself.  Let me know if you see any issues with this please!

str4d:
Thanks for picking this up! I'll review it in the next day or two

cuviper:
Thanks for the counter-review!

bors r+

PR #340, opened 2017-10-17
Fix documentation formatting with commonmark enabled

mbrubeck:
This makes formatting correct with the new pulldown-cmark Markdown parser (rust-lang/rust#44229).


cuviper:
Great, thanks!

bors r+

Issue #233, opened 2016-09-29
Implement std::iter::{Sum,Product}


cuviper:
The [Sum](https://doc.rust-lang.org/std/iter/trait.Sum.html) and [Product](https://doc.rust-lang.org/std/iter/trait.Product.html) traits are stable in 1.12.0, so we could implement them for our types.

(But we consider it a breaking change to increase the required rustc.)

cuviper:
It's only that pesky "breaking change" label.  There is a `next` branch now though, if you'd like to submit a pull request there.  We could also consider it for the stable branch if it used a build script to gate the code, depending on the detected rustc version.

JP-Ellis:
Version 1.12 was released on 21 Oct 2016, so I thought that it might be fine to have a breaking change now (especially given that the crate hasn't reached v1.0 yet), but I understand where you're coming from then.

Issue #347, opened 2017-12-09
Surprising behavior when right-shifting a negative BigInt


mhogrefe:
Right-shifting an primitive integer in Rust (and most other languages AFAIK) is equivalent to dividing by a power of 2 and rounding towards -∞. But right-shifting a BigInt rounds towards 0 instead.

You can see the difference by pasting the following into [https://play.rust-lang.org/](https://play.rust-lang.org/):
```
extern crate num;

fn main() {
    println!("-1 >> 1 == {} (primitive)", -1 >> 1);
    println!("-1 >> 1 == {} (BigInt)", num::BigInt::from(-1) >> 1);
}
```
This prints
```
-1 >> 1 == -1 (primitive)
-1 >> 1 == 0 (BigInt)
```

If this behavior is intentional it should probably be documented and unit-tested.

cuviper:
That's interesting. This behavior is neither an arithmetic nor logical right shift, but sort of a weird hybrid. I don't think this was intentional, just a casual side effect of shifting the internal `BigUint` while preserving the sign.

I worry slightly that changing this would be a breaking change, but I think we can call it just a bug fix. The current behavior is unprecedented in other languages AFAIK, and matching the primitive behavior would be much more useful.

cuviper:
Well sure, GMP implements every conceivable variation. :)

I don't know a way to search for this.  @rust-lang has a "crater" tool that compiles every published crate and some more on GitHub, but this isn't generally available for use.  I suppose we could just poll the users forum and /r/rust.

Generally speaking though, I think we want to match primitive operators in all aspects (except overflow).  That's applying the principle of least surprise, and we could implement alternate forms as other methods if we want to generalize choices.

tspiteri:
GMP does have a definition `#define mpz_div_2exp	mpz_fdiv_q_2exp` in *gmp.h*. It is undocumented and I think it's only there for some compatibility with GMP 1, back when that was the only available variation.

cuviper:
This issue was moved to rust-num/num-bigint#1

PR #348, opened 2017-12-10
bigint: impl algebraic iterator traits

hcpl:
This will make bulk addition and multiplication more convenient for both BigUint and BigInt.


cuviper:
You can see this failed rust 1.8 CI. I'm extremely wary of bumping the minimum rust. *However*, a `build.rs` script could check the version and emit a conditional `cfg` for this, if you'd like to explore that route.

This would fix #233.

hcpl:
Thanks for the idea! I've chosen the `#[cfg(impl_sum_product_for_bigints)]` long name to be explicit about its purpose.

hcpl:
Tests seem to run successfully, but in fact `cfg`-gated tests don't even appear in 1.15.0 build. However they do on beta and nightly CI, and also on my local 1.17.0 one.

hcpl:
Turns out I forgot that picking up `build.rs` by default is a recent addition.

cuviper:
> Tests seem to run successfully, but in fact cfg-gated tests don't even appear in 1.15.0 build.

I'm glad you caught that -- we should figure out a way for CI to catch this automatically, and explicitly make sure that 1.12+ get the expected `cfg`.  Maybe something like `bigint/examples/sum_product.rs` where `main()` panics if the feature is missing, and then make CI try to run this for 1.12 and later.

cuviper:
I might also prefer to create our own version check, rather than holding to an older `rustc_version`, but that can be considered later.

hcpl:
I feel like I need to provide an explanation in case if someone thinks of my code being overly complicated (because it is).

After reading @cuviper's suggestion, I thought it would be better to _embed the checker_ in the main code, which has greater scope of enforcement than examples and even tests. Knowing that something like this may be done at compile-time, I wrote down something like this:

```rust
#[cfg(and(rustc_version_matches(">= 1.12.0"), not(impl_sum_product_for_bigints)))]
compile_error!("\
    cfg option `impl_sum_product_for_bigints` not detected when compiling on \
    rustc {} (this option must be enabled for rustc 1.12.0 onwards)", env!("RUSTC_VERSION"));
```

Unfortunately, there are some blocking issues to this solution:

1. no `rustc_version_matches` to use in `#[cfg]`;
2. `compile_error!` is stable only from 1.20;
3. no `RUSTC_VERSION` environment variable available at compile-time (alternatively, a `RUSTC_VERSION` constant in `std` could be used, though this must be even less likely to exist if `std` wants to retain its API compiler-independent as much as possible).

Instead I've implemented a convoluted solution which checks only at runtime, but is still available when _testing_, not running examples. This one is what you see in the commit before this comment.

Speaking of runtime checks, at first I tried to avoid using generated code in favor of simpler environment variables using `println!("cargo:rustc-env=RUSTC_VERSION={:?}", rustc_version::version())` in build script. However, this is a 1.19+ feature.

So regarding aesthetics, this solution looks ugly to me, but it does the job. Still, I think it could be done better, I just don't know how.

As a side note, I'd like to address issues (1) and (3) ((1) is more important) in the list to Rust devs. Would be very cool if we'll be able to write code similar to one at the top at some point in the future.

cuviper:
I am both amazed and horrified... 😉 

I'm contemplating a semver bump of num-bigint to deal with #347, and if I do that, it's totally reasonable to also raise the bar to *at least* Rust 1.12.

hcpl:
I'd be very glad to see `num` using new Rust features and new versions of libraries that abandoned old Rusts to keep the pace with the ecosystem and ease the maintenance burden. Though if you are inclined to support old Rusts, I'll totally understand that -- there are many reasons when doing so is a good thing.

Still, there is a question about whether the core team will follow the "semver-bump only when something will break" policy after such a bump. If yes, then there ought to be a way to introduce breaking changes while retaining ability to build like they never happened, or else we'll end up collecting technical debt over time when no work to pay said debt is ever done. Also I'm very sure the issue about minimum Rust version has been discussed extensively elsewhere, and we can potentially benefit from that fact too.

cuviper:
The reason I take a hard line on rustc version is because cargo doesn't deal with it.  Cargo will greedily take the newest semver-compatible crate available, but doesn't know if that newer crate can be compiled with your rustc at all.  If crates get the ability to specify this, and cargo uses that information for depsolving, then there will be a strong motivation for me to make a breaking bump to that newly-able version.  Bumping rustc requirements from there on could be handled smoothly without a semver bump.

Yes, this has been discussed elsewhere, and the community seems split on whether keeping such rustc compatibility is worthwhile.  Some are firm like me, some prefer latest-stable only.  There have been a few RFCs towards getting cargo to deal with this, but none have gained enough traction yet.

cuviper:
Please do move this to the new repo and issue, rust-num/num-bigint#4.

(I may hold on merging it for a little bit, until I'm sure I'm done with 0.1.)

PR #349, opened 2017-12-11
[WIP] Make Shr for negative BigInt round down, like primitives do

### LINKED ISSUE #347: Surprising behavior when right-shifting a negative BigInt

cuviper:
Primitive integers always round down when shifting right, but `BigInt`
was effectively rounding toward zero, because it just kept its sign and
used the `BigUint` magnitude rounded down (always toward zero).

Now we adjust the result of shifting negative values, and explicitly
test that it matches the result for primitive integers.

Fixes #347.

### DIFF
--- a/bigint/src/bigint.rs
+++ b/bigint/src/bigint.rs
@@ -227,12 +227,23 @@ impl<'a> Shl<usize> for &'a BigInt {
     }
 }
 
+// Negative values need a rounding adjustment if there are any ones in the
+// bits that are getting shifted out.
+fn shr_round_down(i: &BigInt, rhs: usize) -> bool {
+    i.is_negative()
+        && biguint::trailing_zeros(&i.data)
+            .map(|n| n < rhs)
+            .unwrap_or(false)
+}
+
 impl Shr<usize> for BigInt {
     type Output = BigInt;
 
     #[inline]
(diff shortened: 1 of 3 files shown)


cuviper:
WIP because I want to wait for community feedback [here](https://users.rust-lang.org/t/feedback-requested-on-shr-for-negative-bigint/14361).  If it's deemed too much of a breaking change to call it a bug fix, then I'll re-target this for the `next` branch.

cuviper:
Moved to rust-num/num-bigint#8.

PR #167, opened 2016-02-17
Reorder ops in LCM to avoid overflow.

pnkfelix:
Reorder ops in LCM to avoid overflow.

Fix #166

PR #350, opened 2017-12-14
Avoid large intermediate product in LCM

mhogrefe:
Changed the implementation of BigUint LCM from
`((self * other) / self.gcd(other))`
to
`self / self.gcd(other) * other`

The division is exact in both cases, so the result is the same, but the new code avoids the potentially-large intermediate product, speeding things up and using less memory.

I also removed the unnecessary parentheses, because I think it's clear what order everything will be executed in. But if others think differently I can add them back.

### DIFF
--- a/bigint/src/biguint.rs
+++ b/bigint/src/biguint.rs
@@ -955,7 +955,7 @@ impl Integer for BigUint {
     /// Calculates the Lowest Common Multiple (LCM) of the number and `other`.
     #[inline]
     fn lcm(&self, other: &BigUint) -> BigUint {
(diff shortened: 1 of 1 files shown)


hcpl:
Don't worry, the build is failing because the `syntax` API changed again, which is expected from nightlies. My PR #348 is suffering from this too.

cuviper:
Ugh, num-macros... I'm inclined to finally let that die.

This PR is good though -- we already did similar to avoid overflowing primitives in #167.

cuviper:
#351 nuked num-macros, so let's give this a try:

bors r+

PR #353, opened 2017-12-14
derive: update dependencies

### LINKED ISSUE #352: #[derive(FromPrimitive)] panics on pub(crate)
This does not compile:

```
#[derive(Debug, FromPrimitive)]
pub(crate) enum Foo {
    A = 1,
    B = 2,
}
```

removing `(crate)` makes it work

hcpl:
Fixes #352.

Newer versions support new syntax, such as pub(restricted).

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -18,6 +18,8 @@ extern crate proc_macro;
 use proc_macro::TokenStream;
 
 use syn::Body::Enum;
+use syn::ConstExpr::Lit;
+use syn::Lit::Int;
 use syn::VariantData::Unit;
 
 #[proc_macro_derive(FromPrimitive)]
@@ -42,8 +44,8 @@ pub fn from_primitive(input: TokenStream) -> TokenStream {
                     panic!("`FromPrimitive` can be applied only to unitary enums, {}::{} is either struct or tuple", name, ident)
                 },
             }
-            if let Some(val) = variant.discriminant {
-                idx = val.value;
+            if let Some(Lit(Int(value, _))) = variant.discriminant {
+                idx = value;
             }
             let tt = quote!(#idx => Some(#name::#ident));
             idx += 1;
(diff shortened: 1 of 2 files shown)


cuviper:
```
warning: unreachable expression
  --> derive/tests/trivial.rs:15:43
   |
15 | #[derive(Debug, PartialEq, FromPrimitive, ToPrimitive)]
   |                                           ^^^^^^^^^^^
   |
   = note: #[warn(unreachable_code)] on by default

error[E0004]: non-exhaustive patterns: type Color is non-empty
  --> derive/tests/trivial.rs:15:43
   |
15 | #[derive(Debug, PartialEq, FromPrimitive, ToPrimitive)]
   |                                           ^^^^^^^^^^^
   |

help: Please ensure that all possible cases are being handled; possibly adding wildcards or more match arms.
  --> derive/tests/trivial.rs:15:43
   |
15 | #[derive(Debug, PartialEq, FromPrimitive, ToPrimitive)]
   |                                           ^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `num-derive`.
```

derive/src/lib.rs
```diff
@@ -42,8 +44,8 @@ pub fn from_primitive(input: TokenStream) -> TokenStream {
                     panic!("`FromPrimitive` can be applied only to unitary enums, {}::{} is either struct or tuple", name, ident)
                 },
             }
-            if let Some(val) = variant.discriminant {
-                idx = val.value;
+            if let Some(Lit(Int(value, _))) = variant.discriminant {
```
cuviper: Can we support the full `ConstExpr`, instead of picking it apart?  This will fail on even simple values like `-100` which will map to something like `Unary(Neg, Lit(Int(100)))`.  Nevermind more complicated expressions.  (It seems the old syn just failed to parse it at all.)

I'm thinking instead of `idx`, we store the actual `expr` and an `offset` 0, which we encode as `quote!(... #expr + #offset ...)`.  If the next variant has a discriminant, update `expr` and set `offset` back to 0, otherwise reuse the former `expr` with an incremented `offset`.

cuviper:
Please do move this to the new repo and issue, rust-num/num-derive#2

PR #356, opened 2017-12-19
Move all of the num sub-crates to their own repos

### LINKED ISSUE #354: Break up the mono-repo
Discussions around #347 are making me think we probably _should_ bump num-bigint semver, independently of crates like num-traits that are more often public dependencies.  If we bump bigint, we can also take the chance to make other breaking changes, like raising the minimum rustc for `Sum` and `Product` -- https://github.com/rust-num/num/pull/348#issuecomment-351636694 -- or moving to serde-1 #281.

But it's going to get nasty trying to juggle different compatibility requirements from one shared repo, so I'm now planning to move each crate into its own rust-num/num-* repo.

cuviper:
The separate repos are all in place, so now it's time to trim this one back.

Fixes #354.


cuviper:
I'll need to migrate issues and notify the outstanding PRs too...

cuviper:
Issue and PR updates are done, so let's go...

bors r+

PR #367, opened 2018-06-23
Release 0.2.0

### LINKED ISSUE #298: Missing default-features = false for bigint dependency
When using `num = { default-features = false, features = ["bigint"] }`, `num-bigint` still pulls in a dependency on `rustc-serialize`.

### LINKED ISSUE #357: Update to Serde 1.0
See #281, and this was previously in the next branch after #305.

- [x] rust-num/num-bigint#9
- [x] rust-num/num-complex#8
- [x] rust-num/num-rational#9

### LINKED ISSUE #360: nostd feature
Hey, it'd be nice if this crate could support `nostd`. My use-case is that I'd like to use this crate on a platform that doesn't provide an `std`, but still has an allocator (which I think is the only required bit for most of the crates in here ?). Is this something you'd be willing to accept PRs for ?

cuviper:
Closes #298.
Closes #357.
Closes #360.


BelfordZ:
nice work 💯

